PR #24, opened 2015-11-08
adding coveralls integration

heckj:


.travis.yml
```diff
@@ -5,6 +5,6 @@ node_js:
 services:
  - mongodb
  - rabbitmq
-
-before_install:
- - sudo apt-get update -qq
+after_success:
+ - ./node_modules/.bin/istanbul cover -x "**/spec/**" ./node_modules/.bin/_mocha --report lcovonly -- $(find spec -name '*-spec.js') -R spec --require spec/helper.js
```
benbp: I think it would be better to put this command into the package json as an npm script, so that in the travis file we can just say something like `npm coverage`

heckj: I debated if we should push that into package.json as a script “npm run coverage” or if it made more sense as a standalone script. It’s different from what Jenkins would use (which uses the xunit-file output), so I opted for a direct execution of the specifics  in .travis.yml

benbp:
:+1:

VulpesArtificem:
:+1:

tldavies:
+1

PR #31, opened 2015-11-17
removed globalstrict as it is deprecated

tldavies:
When this is left in it causes jshint warnings when trying to override jshint for a certain file.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -40,7 +40,6 @@
     "evil" : false,
     "expr" : true,     // true: Tolerate `ExpressionStatement` as Programs
     "funcscope" : false,
-    "globalstrict" : false,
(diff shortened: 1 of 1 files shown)


tldavies:
@RackHD/corecommitters

benbp:
:+1:

DavidjohnBlodgett:
@benbp @tldavies should we be ignoring the code coverage issues? Not sure I understand how we reduced it with a delete...

tldavies:
@DavidjohnBlodgett No we should not in most cases. Since the change was in jshintrc though it should not actually be include in code coverage.

DavidjohnBlodgett:
:+1:

PR #34, opened 2015-12-01
MAG-289 [ORFS-97] Add create delete raid graph

WangWinson:
This PR is to add graph for create or delete raid then refresh the MegaRAID catalog
Dependency PR https://github.com/RackHD/on-tasks/pull/76

@RackHD/corecommitters @iceiilin @pengz1 @sunnyqianzhang

### DIFF
--- a/lib/graphs/create-megaraid-graph.js
+++ b/lib/graphs/create-megaraid-graph.js
@@ -1,0 +1,50 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Create RAID via Storcli',
+    injectableName: 'Graph.Raid.Create.MegaRAID',
+    tasks: [
+        {
+            label: 'set-boot-pxe',
+            taskName: 'Task.Obm.Node.PxeBoot',
+            ignoreFailure: true
+        },
+        {
+            label: 'reboot',
+            taskName: 'Task.Obm.Node.Reboot',
+            waitOn: {
+                'set-boot-pxe': 'finished'
+            }
+        },
+        {
+            label: 'bootstrap-ubuntu',
+            taskName: 'Task.Linux.Bootstrap.Ubuntu',
+            waitOn: {
+                'reboot': 'succeeded'
+            }
+        },
+        {
+            label: 'create-raid',
+            taskName: 'Task.Raid.Create.MegaRAID',
+            waitOn: {
+                'bootstrap-ubuntu': 'succeeded'
+            }
+        },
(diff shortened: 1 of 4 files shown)


spec/lib/graphs/delete-megaraid-graph-spec.js
```diff
@@ -0,0 +1,15 @@
+// Copyright 2015, EMC, Inc.
+
+'use strict';
+
+describe(require('path').basename(__filename), function () {
+    var base = require('./base-graph-spec');
+
+    base.before(function (context) {
+        context.taskdefinition = helper.require('/lib/graphs/delete-megaraid-graph.js');
```
WangWinson: Hound doesn't mark global.helper as a lint exception.

lib/graphs/create-megaraid-graph.js
```diff
@@ -0,0 +1,43 @@
...
+        },
+        {
+            label: 'refresh-catalog-megaraid',
+            taskName: 'Task.Catalog.megaraid',
+            waitOn: {
+                'create-raid': 'succeeded'
+            }
+        }
+    ]
+};
```
yyscamper: Usually this kind of workflow should be end with a "reset" task, we don't want the node stays in micro-kernel.

lib/graphs/delete-megaraid-graph.js
```diff
@@ -0,0 +1,43 @@
...
+        },
+        {
+            label: 'refresh-catalog-megaraid',
+            taskName: 'Task.Catalog.megaraid',
+            waitOn: {
+                'delete-raid': 'succeeded'
+            }
+        }
+    ]
+};
```
yyscamper: Usually this kind of workflow should be end with a "reset" task, we don't want the node stays in micro-kernel.

WangWinson: Good point ! Will update

spec/lib/graphs/delete-megaraid-graph-spec.js
benbp: Figured it out, hound is not reading our test .jshintrc in the spec directory https://github.com/thoughtbot/hound/issues/889

benbp:
:+1:

yyscamper:
:+1:

spec/lib/graphs/delete-megaraid-graph-spec.js
WangWinson: Thanks @benbp. Good to know this kind of info

PR #37, opened 2015-12-04
GRIP-22: Add support for booting LiveCD

zyoung51:


heckj:
@zyoung51 cool - which LiveCD are you testing this out with?

zyoung51:
SIO provided their test image which is a RHEL based LiveCD.  The setup_iso.py (on-http/data/templates/setup_iso.py) is used to convert that ISO image to the artifacts that are booted with these PRs.

zyoung51:
Unit test failure due to missing dependency: https://github.com/RackHD/on-tasks/pull/87.  The graph references a task that doesn't exist yet.

heckj:
Ah, very cool! With the vagrant work and your ISO unpack script setup, I was looking to snag a public one to try out and vet an install process :^)

lib/graphs/boot-livecd-graph.js
```diff
@@ -0,0 +1,34 @@
...
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Boot LiveCD',
+    injectableName: 'Graph.BootLiveCD',
+    options: {
+        defaults: {
+            repo: '{{api.server}}/LiveCD/{{options.version}}'
+        },
```
yyscamper: For ODR-273, we need to define a timeout value, just like we did for install-esx and install-rhel.
https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/install-esx-graph.js#L15

yyscamper:
@zyoung51 , do we still need manually run the script "setup_iso.py" to ensure the LiveCD workflow success?

jlongever:
+1

lib/graphs/boot-livecd-graph.js
yyscamper: Just as I commented in this PR (https://github.com/RackHD/on-tasks/pull/87), if user forgets to add the completionUrl query in the LiveCD, the workflow will hang. So I recommend to add a reasonable timeout here so even the workflow is hang, it will be hang in acceptable duration.

yyscamper:
+1 after adding the timeout

PR #38, opened 2015-12-07
Deploy debian packages to bintray

jlongever:
@RackHD/corecommitters

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,14 @@ matrix:
   allow_failures:
    - node_js: "4"
 
+addons:
+  apt:
+    packages:
+      - dh-make
+      - devscripts
+      - debhelper
+      - git
+
 services:
  - mongodb
  - rabbitmq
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,6 +24,25 @@ after_success:
...
+ - git log -n 1 --pretty=format:%h.%ai.%s > commitstring.txt
+ - debuild --no-lintian --no-tgz-check -us -uc
+ - popd
+ - cp -a *.deb deb/
+
+deploy: 
+  provider: bintray
+  file: .bintray.json
+  user: $BINTRAY_USER
+  key: $BINTRAY_KEY
```
yyscamper: Where is the USRER/KEY from?

heckj: These are provided through the travis ENV variables set up per repository. For an example, you can see the configuration control (as an admin on the repo) for the `on-imagebuilder` repo at https://travis-ci.org/RackHD/on-imagebuilder/settings.

heckj:
we probably want to manually update the debian changelog string to have a more recently version than the defaults in there now (which are legacy from months ago), or do a similar `dch` command for force a local version to be clear on when it was built.

Also - will this build on each PR, or will it only build when merged into master?

jlongever:
@heckj should only deploy when changes are committed to the master branch.

jlongever:
@heckj I'll update to run dch.

.travis.yml
jlongever: They are set as travis-ci variables.

jlongever:
@heckj for the moment i'll run dch to update the changelog with the branch (should just be master) and commit log.  I assume we don't want to deploy new .debs for every master commit?

heckj:
Works for me!

benbp:
:+1:

zyoung51:
+1

PR #39, opened 2015-12-21
rationalizing logs

heckj:
updating mapping to more constricted set of logs per
https://github.com/RackHD/RackHD/issues/34

**NOTE**: Requires https://github.com/RackHD/on-core/pull/48 to be
committed prior as it adds the synonym 'critical' to the existing 'crit'
level.

### DIFF
--- a/index.js
+++ b/index.js
@@ -30,9 +30,12 @@ taskGraphRunner.start()
         logger.info('Task Graph Runner Started.');
     })
     .catch(function(error) {
+        //NOTE(heckj): I'm unclear why this is on the console directly and not
+        // using a logger. Would expect this to be logger.critical(), but
+        // leaving as is because I don't know the implications.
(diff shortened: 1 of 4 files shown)


heckj:
@benbp when I was reviewing this code set, the GraphContext (https://github.com/RackHD/on-taskgraph/blob/master/lib/context.js#L61) had a curried log function with all internal log messages set to 'info', which seemed reasonably close, but I wondered if that wasn't more appropriate to debug than info.

Since you're in a big refactor, I didn't want to mess with anything in there deeply, but wanted to ask you if `info` was appropriate or not.

benbp:
It's pretty noisy so debug is probably right. :+1:

DavidjohnBlodgett:
:+1:

yyscamper:
:+1:

PR #43, opened 2016-01-13
Set mgmt server discovery task in serial manner

iceiilin:
Set mgmt server discovery tasks work in serial manner to avoid task scheduling competition. Otherwise a mongoError is likely to report with a long dump log:
2016-01-12T09:49:55.806Z [on-taskgraph] [TaskGraph.TaskGraph] [Server] Error persisting Task Graph state
 -> /lib/task-graph.js:483     
    """
      Details:  { name: 'MongoError',
        message: 'Object #<Object> has no method \'value\'',
        driver: true,
        ok: 1,
        n: 0,
        code: 14,
        errmsg: 'Object #<Object> has no method \'value\'',
        writeErrors: 
         [ { index: 0,
             code: 14,
             errmsg: 'Object #<Object> has no method \'value\'' } ] }
      
    """

### DIFF
--- a/lib/graphs/discovery-mgmt-graph.js
+++ b/lib/graphs/discovery-mgmt-graph.js
@@ -14,11 +14,17 @@ module.exports = {
         {
             label: 'catalog-mgmt-lldp',
             taskName: 'Task.Catalog.Local.LLDP',
+            waitOn: {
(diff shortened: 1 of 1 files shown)


yyscamper:
:+1:

heckj:
:+1:

PR #46, opened 2016-01-15
Bugfix/use ipmitool 1.8.13 1 1.1

johren:
Need to specify the ipmitool version to avoid a current bug in version 1.8.13-1ubuntu0.6 of ipmitool that breaks the fru command. This change should be merged with the similar PRs in onrack-infra onrack-base, and on-taskgraph master branch.


johren:
Opened against master rather than 1.1.0 branch.

PR #56, opened 2016-02-03
Workflow engine refactor

benbp:
Refactoring the workflow engine for:

- High Availability
- Fault Tolerance
- Better performance
- Better execution strategy

This PR completely overhauls the on-taskgraph repository:

- TaskGraph is refactored out into on-core
- Adding TaskScheduler and TaskRunner classes that do the main graph scheduling and task running work, as well as a couple database poller classes.
- Removing all old pieces related to previous TaskGraph execution model: scheduler.js, task-graph-subscriptions.js, registry.js, etc.
- Adding multiple run modes (detailed in the README updates), so it can run in scheduler only mode, task runner only mode, or hybrid mode depending on a user's performance and reliability needs.

Relies on https://github.com/RackHD/on-tasks/pull/116, https://github.com/RackHD/on-core/pull/75, and https://github.com/RackHD/on-http/pull/112

High level architecture and design goals detailed are in the attached slides:
[RackHD Workflow Engine Refactor.pdf](https://github.com/RackHD/on-taskgraph/files/114831/RackHD.Workflow.Engine.Refactor.pdf)


benbp:
Mentioning a mix of those who came to the presentation/should review this: @RackHD/corecommitters @VulpesArtificem @heckj @king-jam @johren @stuart-stanley @jimturnquist @larry-dean @davequick @zyoung51 @zhangy42 @mattgam @tldavies @DavidjohnBlodgett @rolandpoulter

README.md
```diff
@@ -1,19 +1,83 @@
...
+
+*It is highly recommended to create indexes in mongo to reduce the CPU footprint and increase performance*
+
+```
+$ mongo pxe
+> db.createCollection('taskdependencies')
+> db.createCollection('graphobjects')
+> db.taskdependencies.createIndex({ taskId: 1 })
+> db.taskdependencies.createIndex({ graphId: 1 })
+> db.taskdependencies.createIndex({ graphId: 1, state: 1 })
```
rolandpoulter: Couldn't we create these indexes automatically?

benbp: Good point, with waterline native mongo methods we can probably do that.

benbp: I'll make the update to on-core.

lib/completed-task-poller.js
```diff
@@ -0,0 +1,268 @@
...
+     * @param {Object} options
+     * @constructor CompletedTaskPoller
+     */
+    function CompletedTaskPoller(domain, options) {
+        options = options || {};
+        assert.string(domain);
+        this.running = false;
+        this.pollInterval = options.pollInterval || (1000);
+        this.concurrentCounter = { count: 0, max: 1 };
+        this.completedTaskBatchSize = 200;
```
heckj: does it make sense (is it worthwhile) to pull `debug` && `concurrentCounter` from the options stanza, using defaults if not set?

lib/completed-task-poller.js
```diff
@@ -0,0 +1,268 @@
...
+    /**
+     * Poll the store for completed tasks.
+     *
+     * @memberOf CompletedTaskPoller
+     */
+    CompletedTaskPoller.prototype.pollTasks = function() {
+        var self = this;
+        assert.ok(self.running, 'lease expiration poller is running');
+
+        Rx.Observable.interval(self.pollInterval)
```
heckj: request from non-Rx knowledgable monkey: can you toss in some description in the code itself, like you do at line 68, with what these RX chains are doing and intending to do? Some of it is self-explanatory, but I don't entirely grok the intended logic in the chaining of `.map()` after the `.takeWhile` and ending with the `subscribe`

lib/completed-task-poller.js
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    /**
+     * Poll the store for completed tasks.
+     *
+     * @memberOf CompletedTaskPoller
+     */
+    CompletedTaskPoller.prototype.pollTasks = function() {
+        var self = this;
+        assert.ok(self.running, 'lease expiration poller is running');
```
heckj: is this duplicate of the logic in the `takeWhile` statement in line 65? Did you want to intentionally explode on a failed precondition?

lib/completed-task-poller.js
```diff
@@ -0,0 +1,268 @@
...
+
+        Rx.Observable.interval(self.pollInterval)
+        .takeWhile(self.isRunning.bind(self))
+        .map(self.processCompletedTasks.bind(self, self.completedTaskBatchSize))
+        .mergeLossy(self.concurrentCounter)
+        // Don't let processCompletedTasks return a waterline object to the logger
+        // otherwise it will exceed the call stack trying to traverse a circular object
+        .map(null)
+        .subscribe(
+            self.handleStreamDebug.bind(self, 'CompletedTaskPoller stream pipeline success'),
```
heckj: is this relevant to bind around `if (self.debug)`, to only invoke on debug, or intentionally always bound? (not sure of the pattern and what's implied with `handleStreamdebug.bind()`)

lib/completed-task-poller.js
```diff
@@ -0,0 +1,268 @@
...
+     * @param {Object} data
+     * @memberOf CompletedTaskPoller
+     */
+    CompletedTaskPoller.prototype.handlePotentialFinishedGraph = function(data) {
+        assert.object(data);
+        var self = this;
+
+        return Rx.Observable.just(data)
+        .flatMap(function(data) {
+            if (_.contains(Constants.FailedTaskStates, data.state)) {
```
heckj: is `data` object guaranteed to include `data.state` - you only assert on `data` being an object earlier. Sufficient because of calling convention?

lib/completed-task-poller.js
benbp: Yes! I did this for task-scheduler but forgot for this one.

lib/completed-task-poller.js
```diff
@@ -0,0 +1,268 @@
...
+                data.failed = true;
+                data.done = true;
+                return Rx.Observable.just(data);
+            }
+            return store.checkGraphFinished(data);
```
heckj: is this kosher? Reading http://reactivex.io/documentation/operators/flatmap.html is seems to want a single observable back, and wasn't sure what store.checkGraphFinished(data) returned...

lib/service-graph.js
```diff
@@ -9,72 +8,119 @@ module.exports = ServiceGraph;
...
-
-            createAndRunServiceGraph(graph.definition, graph.definition.options)
-            .catch(function(error) {
-                logger.error('Failed to restart service graph ' + graph.definition.injectableName, {
-                    graphId: graph.instanceId,
-                    injectableName: graph.definition.injectableName,
-                    error: error
-                });
-            });
+    exports.createAndRunServiceGraph = function(definition, domain) {
```
heckj: a little JSDoc on this guy would be nice - define what definition and domain are intended to be, as well as that this is returning a promise/chain...

lib/completed-task-poller.js
benbp: So, unlike with a Promise chain, where you would expect a single path of execution and then done, this `Rx.Observable.interval` call produces a continuously running pipeline of the chained calls below, that triggers on every `pollInterval` period. The reason we do `.takeWhile` at the top is to basically implement an auto-disposal mechanism: before doing any of the subsequent logic, first check if we're still running and just shut down whole stream if we're not. That way we don't have to do any asynchronous disposal when stopping the service, we just set a variable `this.running = false`.

The `.map` call is where we actually do the work, and the rest of the stuff is just coordination about when to do it.

Explaining the subscribe will basically be an intro to how an Rx Observable is constructed. The `subscribe` call essentially acts as an implementation of an _Observer_, and it takes three optional callbacks as arguments: `.subscribe(<SUCCESS>, <ERROR>, <COMPLETE>)`. 

The first callback <SUCCESS> is triggered *every* time this observable stream processes all the way through, in this case at most on every pollInterval when there are no failures within the stream. 

The second callback <ERROR> is triggered if there is a stream error, i.e. if a functional call within it bubbles up an error. In this case we make sure that no errors bubble up because we don't want this main interval poller to stop execution on one-off/single failures.

The third callback <COMPLETED> is triggered when the stream is done. In this specific stream this will only be called in an error case, or if `takeWhile` is given a falsey value (`this.running = false`). In streams with fixed data inputs, then <COMPLETED> can also be called on <SUCCESS> cases (e.g. in the case of `Rx.Observable.from([1,2,3])`, we know there is a fixed length to the stream input and can tell when we're done).

A little more on the behavior of `subscribe` here. This observable is what we call a cold observable. This means that the Observable object won't actually start pumping out events on the poll interval until we actually call subscribe (i.e. add an Observer). There is another class of observables called hot observables. These ones will emit events regardless of whether there is an Observer is attached, and the events will just be dropped if there isn't one.

lib/completed-task-poller.js
benbp: In this case it means, before we even set up the stream in the first place, make sure we're running, otherwise we'll create a stream that will never run and immediately complete. Basically some defensive programming to try to prevent accidents where the startup code is executed in the wrong order (e.g. `pollTasks()` _then_ `this.running = true`, that would be buggy because `pollTasks()` would stop execution before `this.running = true` was set).

lib/task-graph-runner.js
```diff
@@ -5,45 +5,85 @@
...
     }
 
-    function stop() {
-        return scheduler.stop().then(function() {
-            return serviceGraph.stop();
-        }).then(function() {
-            return taskGraphSubscriptions.stop();
-        }).then(function() {
-            return core.stop();
+    Runner.prototype.start = function(options) {
```
heckj: a little JSDoc on this baby, especially with what should/could be in `options` would be handy for future reference

lib/completed-task-poller.js
benbp: We do the `if (self.debug)` in the `handleStreamDebug` method itself, because as a general style rule because it saved more lines of code and duplication when used many times for different streams (mostly in other modules). It could just as easily have been expressed as:

```
.subscribe(
    function(data) {
        if (self.debug) {
            self.handleStreamDebug('CompletedTaskPoller stream pipeline success', data);
        }
    }
);
```

lib/completed-task-poller.js
benbp: No there could be a potential bug here, you're right. I'll add another assert statement.

lib/completed-task-poller.js
benbp: `flatMap` also coerces Promise objects, so we just use it for all async calls whether they return Observables or Promises. Ideally, everything underneath would also be Observable returning, but we didn't want to sail too far out into the Rx ocean without testing the shallow waters first and marshaling acceptance and understanding from others.

lib/task-runner.js
```diff
@@ -0,0 +1,410 @@
...
+            .filter(function(task) { return !_.isEmpty(task); })
+            .tap(function(task) {
+                logger.info('Cancelling task', {data: task.toJSON()});
+            })
+            .map(function(task) { return task.cancel(); })
+            .finally(function() {
+                delete self.activeTasks[data.taskId];
+            });
+    };
+    /**
```
heckj: stupid, but reading downward, it took me a while to find this, although it's used in line 128 above. It might be easier to read if you toss this at the top of the file for definitions...

lib/service-graph.js
benbp: Will do.

lib/task-graph-runner.js
benbp: Will do.

lib/task-runner.js
```diff
@@ -0,0 +1,410 @@
...
+                return Task.create(
+                    _data.task,
+                    { instanceId: _data.task.instanceId },
+                    _data.context
+                );
+            })
+            .tap(function(task) {
+                self.activeTasks[task.instanceId] = task;
+            })
+            .tap(function(task) {
```
heckj: is there a reason to separate these `.tap` functions?

lib/task-runner.js
benbp: Good idea. You want us to move the `handleStreamError`, `handleStreamSuccess` and `handleStreamDebug` methods used throughout the various modules to the top as well? Or is it more that from the naming it was unclear what the function was doing without seeing the definition? (cc @VulpesArtificem )

lib/task-scheduler.js
```diff
@@ -0,0 +1,641 @@
...
+        options = options || {};
+        this.running = false;
+        this.schedulerId = options.schedulerId || uuid.v4();
+        this.domain = options.domain || Constants.DefaultTaskDomain;
+        this.evaluateTaskStream = new Rx.Subject();
+        this.evaluateGraphStream = new Rx.Subject();
+        this.checkGraphFinishedStream = new Rx.Subject();
+        this.pollInterval = options.pollInterval || 500;
+        this.concurrencyMaximums = this.getConcurrencyMaximums(options.concurrent);
+        this.findUnevaluatedTasksLimit = 200;
```
heckj: should this be overridable in `options` or intentionally locked down?

lib/task-scheduler.js
```diff
@@ -0,0 +1,641 @@
...
+        this.subscriptions = [];
+        this.leasePoller = null;
+        this.debug = false;
```
heckj: should this be overridable with an `options.debug`?

lib/task-runner.js
heckj: those I grokked more easily, because you had them earlier in this PR and use them a bit more frequently, I just didn't spot `safeStream` as quickly

lib/task-runner.js
benbp: Nope, probably there was stuff in between before some refactoring event and it just got overlooked. Will consolidate.

lib/rx-mixins.js
```diff
@@ -0,0 +1,167 @@
...
+            if (!this.isStopped) {
+                this.isStopped = true;
+                this.o.onError(e);
+                return true;
+            }
+
+            return false;
+        };
+
+        function InnerObserver(parent, sad) {
```
rolandpoulter: Don't you just wish you could use es6 classes for this. :P

lib/task-scheduler.js
benbp: Nope, will update, same for the options.debug suggestion below.

lib/task-scheduler.js
```diff
@@ -0,0 +1,641 @@
...
+     */
+    TaskScheduler.prototype.createTasksToScheduleSubscription = function(evaluateGraphStream) {
+        var self = this;
+        return evaluateGraphStream
+        .takeWhile(self.isRunning.bind(self))
+        .map(self.findReadyTasks.bind(self))
+        .mergeLossy(self.concurrencyMaximums.findReadyTasks)
+        .filter(function(data) { return !_.isEmpty(data.tasks); })
+        .pluck('tasks')
+        .flatMap(function(tasks) { return Rx.Observable.from(tasks); })
```
heckj: Mostly I got this sequence, but this line stumped me. What's the `flatMap` with the `Rx.Observable.from` doing for us here?

lib/task-scheduler.js
```diff
@@ -0,0 +1,641 @@
...
+            evaluateGraphStream.onNext({ graphId: data.graphId });
+        }
+    };
+
+    /**
+     * Stream handler that finds tasks that are ready to schedule and schedules
+     * them.
+     *
+     * @param {Object} evaluateGraphStream
+     * @memberOf TaskScheduler
```
heckj: could you annotate what this returns in the JSDoc? Pretty sure it's an observable, but wasn't entirely certain...

lib/task-scheduler.js
```diff
@@ -0,0 +1,641 @@
...
+    TaskScheduler.prototype.checkGraphSucceeded = function(data) {
+        assert.object(data, 'graph data object');
+        var self = this;
+
+        return Rx.Observable.just(data)
+        .flatMap(store.checkGraphFinished.bind(store))
+        .filter(function(_data) { return _data.done; })
+        .flatMap(store.setGraphDone.bind(store, Constants.TaskStates.Succeeded))
+        .filter(function(graph) { return !_.isEmpty(graph); })
+        .map(function(graph) { return _.pick(graph, ['instanceId', '_status']); })
```
heckj: why the `map` here? Just filtering down the data to the key pieces you care about: `instanceId` and `_status`?

lib/task-scheduler.js
```diff
@@ -0,0 +1,641 @@
...
+     */
+    TaskScheduler.prototype.updateTaskDependencies = function(data) {
+        assert.object(data, 'task dependency object');
+        return Rx.Observable.forkJoin([
+            store.setTaskStateInGraph(data),
+            store.updateDependentTasks(data),
+            store.updateUnreachableTasks(data)
+        ])
+        .flatMap(store.markTaskEvaluated.bind(store, data))
+        .catch(this.handleStreamError.bind(this, 'Error updating task dependencies'));
```
heckj: another dummy Rx question - when you end with the .catch, what's this return? Is it still an `Rx.Observable` ?

lib/task-scheduler.js
```diff
@@ -0,0 +1,641 @@
...
+     *
+     * @param {String} domain
+     * @memberOf TaskScheduler
+     */
+    TaskScheduler.prototype.findUnevaluatedTasks = function(domain) {
+        return Rx.Observable.just()
+        .flatMap(store.findUnevaluatedTasks.bind(store, domain, this.findUnevaluatedTasksLimit))
+        .tap(function(tasks) {
+            if (tasks && tasks.length) {
+                logger.debug('Poller is triggering unevaluated tasks to be evaluated', {
```
heckj: is this log level really `debug` or should it be `warning`?

lib/task-runner.js
benbp: Ah ok.

lib/task-scheduler.js
```diff
@@ -0,0 +1,641 @@
...
+            self.initializePipeline();
+            self.leasePoller = LeaseExpirationPoller.create(self, {});
+            self.leasePoller.start();
+            return [
+                self.subscribeRunTaskGraph(),
+                self.subscribeTaskFinished(),
+                self.subscribeCancelGraph()
+            ];
+        })
+        .spread(function(runTaskGraphSubscription, taskFinishedSubscription) {
```
heckj: you return a trio of data, but only use two of them. Intentional?

spec/lib/completed-task-poller-spec.js
```diff
@@ -0,0 +1,326 @@
...
+        };
+    };
+
+    before(function() {
+        helper.setupInjector([
+            helper.require('/lib/completed-task-poller.js'),
+            core.workflowInjectables
+        ]);
+        Rx = helper.injector.get('Rx');
+        //Promise = helper.injector.get('Promise');
```
heckj: remove line instead of leaving it commented out?

lib/rx-mixins.js
benbp: Lol, no why would I possibly want to write cleaner code?

lib/task-scheduler.js
benbp: Ideally I would just do `.pluck('tasks').from()` but that didn't seem to work so I had to flatMap the observable.

What it's actually doing is taking a single stream event with an array data type, and emitting multiple new stream events, one for each element in the array, that are all non-blocking on each other, because we want to process `handleScheduleTaskEvent` for each task independently of the others, but it's better for performance to get them all at once with a single database query.

lib/task-scheduler.js
benbp: Yep, will do, and you're right, it returns an Observable.

lib/task-scheduler.js
heckj: ^^ would you toss that into an inline comment in the code? That detail would be great to persist beyond this PR review...

lib/task-scheduler.js
benbp: Correct. We prune data in a lot of places to make it easier to track what values are need vs. what are noise, and also defensively so we don't pass huge objects into logging methods to get printed out.

lib/task-scheduler.js
benbp: Yes, though `.catch` actually returns a _new_ Observable instance that is separate from the current one we're in at this point. In the case of `handleStreamError` we return an `Rx.Observable.empty()` object. This enables us to handle errors and prevent them from bubbling up to the outer Observable stream so it doesn't stop execution with a call to `onError`. It also prevents the outer stream from continuing beyond this call, which we don't want. So basically: stop execution of this instance of the outer stream pipeline right at this point, but don't cause any failures to it either.

lib/task-scheduler.js
benbp: This is a common operation under load. While sometimes it indicates there were dropped events over AMQP or a previous processing task was interrupted by a crash, it more commonly happens when we drop things in our queue because of high load to be picked up later during the next poll.

lib/rx-mixins.js
```diff
@@ -0,0 +1,167 @@
...
+
+function rxMixins(
+    Rx,
+    assert
+) {
+
+    var MergeLossyObserver = (function () {
+        function MergeLossyObserver(o, concurrentCounter, g) {
+            this.o = o;
+            this.g = g;
```
heckj: reading the spec that tests this, we don't pass in anything for `g` - what's `g` doing and why are adding it to `sad` in `handleSubscribe`. That should probably be obvious to me, but I'm missing this one...

lib/task-scheduler.js
benbp: I think `info` would also be too noisy, because under high load this will be happening for 100+ tasks (or whatever the limit is specified has) every poll interval (500ms by default). So that's waaaay to much logging at info level IMO.

lib/task-scheduler.js
benbp: Nope, last minute addition and we missed it. @VulpesArtificem FYI.

spec/lib/completed-task-poller-spec.js
benbp: Whoops!

lib/task-scheduler.js
benbp: Absolutely.

spec/lib/task-runner-spec.js
```diff
@@ -0,0 +1,515 @@
...
+            try {
+                cb(data);
+                done();
+            } catch (e) {
+                done(e);
+            }
+        };
+    };
+
+    var streamOnCompletedWrapper = function(stream, done, cb) {
```
heckj: a little JSDoc comment about what this is doing would make the tests a bit easier to read.

lib/rx-mixins.js
benbp: Don't worry, I don't understand it very well either. This is mostly hacked together code I grabbed from the throttle implementation out of the Rx.JS source (hence the Microsoft Apache license up top).

The `sad` is a SingleAssignmentDisposable, basically we're many levels of nested observable instances deep at this point, and I think the `sad` ensures that one disposable event disposes all of these instances at once. The `g` is created below with `var g = new Rx.CompositeDisposable();` which to the best of my understanding is also part of this shared disposer handling.

heckj:
phew, that was a deep one :+1: with the relevant updates you said you'd make in the comments above

spec/lib/task-runner-spec.js
benbp: Will do, and I'll add it here first:

This is the way we're testing all of our inner stream creation functions. Since they all return un-subscribed cold Observable objects, we can hijack the subscribe method to one we control in the unit test. Since this is asynchronous, we have to use the Mocha/Chai library's `done` callback, so basically what we're doing is adding a helper method that does:

Subscribe to the stream completed event (thus starting it), and _on_ that event run a callback that includes our test assertions, and calls the `done` callback appropriately so that we can pass or fail the test.

benbp:
Ha! Congrats on making it through alive, and triple thanks for the review!

lib/task-scheduler.js
stuart-stanley: I believe this is also a recovery point for if someone say, reboots the box we are running from? One path being that on restart all the tasks that were in the middle of evaluation are gone? The other being the more similar-to-overload case of all the AMQP messages getting dropped in the restart.  (Asking more to understand than poke at the code)

benbp: Correct! Any tasks that were in the middle of running right before the reboot will be started from the beginning, and the rest of the graph state will be what it was. This was also partially a reason why we set up discovery to always have the bootstrapper task running in parallel, so if the node or workflow runner restarts that task will still be treated as active to pxe boot nodes.

One minor weakness I should mention, that we plan on fixing, is that task timeouts are stored in memory and not in the database right now, so if you have a 2 hour timeout, for example, and the process crashes for 2 hours, we'll still try to run that task for 2 more hours, rather than failing it outright on restart.

benbp: @stuart-stanley one correction to what you said though: if a task state is finished in the database but we haven't completed evaluation, we don't lose that information (i.e. the task is still treated as finished). We enforce a database transaction as part of this call so that we just re-evaluate all of the tasks depending on the current one to make sure we didn't miss anything because of the crash/reboot.

lib/task-runner.js
```diff
@@ -0,0 +1,410 @@
...
+    };
+    /**
+     * Wraps promise or observable returning functions in their own new observable
+     * sequence to catch failures
+     * @memberOf TaskRunner
+     * @param {Object} toObserve - Promise/Observable to wrap
+     * @param {String} msg - an error message for if the wrapped observable fails
+     * @param {Object} the data to be mapped with the given Promise/Observable
+     */
+    var safeStream = function(toObserve, msg, streamData) {
```
rolandpoulter: Is there a reason you're not using the function literal form?
```function safeStream(...) {```

I'm not sure if they fixed this in node/v8, but an anonymous function in a call stack isn't as convenient as one with a literal name. They might infer the name now, in which case disregard.

rolandpoulter: I think they only infer the name if its a prototype method.

benbp: Agreed, we will change this (attn: @VulpesArtificem ).

jlongever:
+1 assuming those failures are due to missing dependent PR's

anhou:
+1

PR #57, opened 2016-02-06
Update Dockerfile for Alpine Linux

rolandpoulter:
REQUIRES: https://github.com/RackHD/on-core/pull/79
* Switched rackhd/on-core to use an alpine linux base image.
* Use CMD instead of ENTRYPOINT for easier debugging.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+.bintray.json
+.dockerignore
+.git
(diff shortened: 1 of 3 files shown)


benbp:
:+1:

VulpesArtificem:
:+1:

PR #58, opened 2016-02-12
ODR-547 Updated discovery-refresh-graphs.js to use sku discovery

jimturnquist:
Updated discovery-refresh-graphs.js to use sku discovery tasks instead of just discovery.

### DIFF
--- a/lib/graphs/discovery-refresh-graph.js
+++ b/lib/graphs/discovery-refresh-graph.js
@@ -5,25 +5,39 @@
 module.exports = {
     friendlyName: 'Refresh node',
     injectableName: 'Graph.Refresh.Discovery',
-    "options": {
-        "discovery-refresh-graph": {
-            "graphOptions": {
-                "target": null
-            }
+    options: {
+        'reset-at-start': {
+            nodeId: null
         },
-        "reboot-at-start": {
-            "nodeId": null
-        }
+        'discovery-refresh-graph': {
+            graphOptions: {
+                target: null
+            },
+            nodeId: null
+        },
+        'generate-sku': {
+            nodeId: null
+        },
+        'generate-enclosure': {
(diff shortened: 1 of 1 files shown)


heckj:
If you run this with pollers already existing, will it create another set of pollers?

benbp:
I believe it will: https://github.com/RackHD/on-tasks/blob/master/lib/jobs/create-default-pollers.js#L68-L75

jimturnquist:
@Ben:   Would it be possible to test if pollers are already enabled and just skip that.  If a discovery-refresh-graph is in play, it really can't assume where the previous discovery stopped short.  Any thoughts?

benbp:
@jimturnquist yeah, the easiest way to me seems to be updating the workitems.find call to the database to be a workitems.findOrCreate call (see https://github.com/RackHD/on-tasks/blob/master/lib/jobs/create-default-pollers.js#L74).

Some example usage of findOrCreate: https://github.com/RackHD/on-core/blob/ea389bb25bdd6c890880442730481d78ac6ad052/lib/services/environment.js#L32-L35

Basically, `findOrCreate(query, document)`. You'd want the query to include the config in the query I believe (i.e. `query = { node: self.nodeId, config: { command: <cmd> } }`), since it gets called multiple times as it iterates through the poller options.

jimturnquist:
Thanks Ben.

jimturnquist:
@benbp @heckj I was able to make the on-tasks changes suggested by Ben to use findOrCreate.  Tests have been updated as well.

yyscamper:
:+1:

johren:
@RackHD/corecommitters

jlongever:
test this please

jlongever:
^^ ignorable smoke test failures.
+1

PR #59, opened 2016-02-12
ODR-547 Updated discovery-refresh-graphs.js to use sku discovery

jimturnquist:
ODR-547 Updated discovery-refresh-graphs.js to use sku discovery tasks instead of just discovery.

(cherry picked from commit 90928e38ff98a2a4d2880233f91dbf4adabfef3d)

### DIFF
--- a/lib/graphs/discovery-refresh-graph.js
+++ b/lib/graphs/discovery-refresh-graph.js
@@ -5,15 +5,27 @@
 module.exports = {
     friendlyName: 'Refresh node',
     injectableName: 'Graph.Refresh.Discovery',
-    "options": {
-        "discovery-refresh-graph": {
-            "graphOptions": {
-                "target": null
-            }
+    options: {
+        'reboot-at-start': {
+            nodeId: null
         },
-        "reboot-at-start": {
-            "nodeId": null
-        }
+        'discovery-refresh-graph': {
+            graphOptions: {
+                target: null
+            },
+        },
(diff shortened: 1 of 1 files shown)


lib/graphs/discovery-refresh-graph.js
```diff
@@ -34,7 +46,88 @@ module.exports = {
...
+                'discovery-refresh-graph': 'succeeded'
+            },
+            taskName: 'Task.Catalog.GenerateEnclosure',
+            ignoreFailure: true
+        },
+        {
+            label: 'create-default-pollers',
+            taskDefinition: {
+                friendlyName: 'Create Default Pollers',
+                injectableName: 'Task.Inline.Pollers.CreateDefault',
```
yyscamper: Though I know both [Graph.MgmtSKU.Discovery](https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/discovery-mgmtsku-graph.js) and [Graph.SKU.Discovery](https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/discovery-sku-graph.js) use the inline task definition to create default pollers, but I see we have a task definition [Task.Pollers.CreateDefault](https://github.com/RackHD/on-tasks/blob/master/lib/task-data/tasks/create-default-poller.js), so I think it's better to leverage the existing task.

yyscamper:
:+1: though a minor suggestion.

PR #61, opened 2016-02-17
Odr 547 master

jimturnquist:
master pull request for ODR-547 work.

### DIFF
--- a/lib/graphs/discovery-refresh-delayed-graph.js
+++ b/lib/graphs/discovery-refresh-delayed-graph.js
@@ -3,12 +3,9 @@
 'use strict';
 
 module.exports = {
-    friendlyName: 'Refresh node',
-    injectableName: 'Graph.Refresh.Discovery',
+    friendlyName: 'Refresh node delayed',
+    injectableName: 'Graph.Refresh.Delayed.Discovery',
     options: {
-        'reset-at-start': {
-            nodeId: null
-        },
         'discovery-refresh-graph': {
             graphOptions: {
                 target: null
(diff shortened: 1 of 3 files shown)


lib/graphs/discovery-refresh-delayed-graph.js
```diff
@@ -0,0 +1,111 @@
...
+        },
+        {
+            label: 'generate-sku',
+            waitOn: {
+                'discovery-refresh-graph': 'succeeded'
+            },
+            taskName: 'Task.Catalog.GenerateSku',
+        },
+        {
+            label: 'generate-enclosure',
```
heckj: double checking - is this task idempotent, or will you get multiple enclosures by repeatedly running this?

heckj: (looks like if it fails if one already exists, we're safe with `ignoreFailure: true` just below)

heckj:
Question about the enclosures (should have spotted that earlier), but otherwise :+1: if you're good with the ramifications. Side note to be clear, if the delayed graph is invoked, the workflow will be "in existence" until the node reboots and does that process, which means other workflows may not be able to be run against it - in particular, I don't know that you'll be able to invoke a reboot workflow against that same node...

jimturnquist:
@heckj Yep.  The delayed case will work exactly as you mentioned.  I believe the use case is when a node is taken out of service and then added back in.  The node id should "stick" but be refreshed on reboot.

lib/graphs/discovery-refresh-delayed-graph.js
iceiilin: If the enclosure relation information already exists, the job will finish successfully without changing database. So it is fine here

jimturnquist:
@RackHD/corecommitters I noticed this was still hanging out there.  I wasn't sure if there was something being waited on in regards to this.  The release 1.1 version is already in.

jlongever:
+1  
just waiting for some jenkin's CI cleanup wrt recent workflow engine commits.

benbp:
:+1:

benbp:
test this please

PR #62, opened 2016-02-17
Add node tagging tasks to graphs

zyoung51:
@RackHD/corecommitters @uppalk1 @keedya @tannoa2 
Depends on https://github.com/RackHD/on-tasks/pull/124, https://github.com/RackHD/on-core/pull/88

### DIFF
--- a/lib/graphs/discovery-mgmtsku-graph.js
+++ b/lib/graphs/discovery-mgmtsku-graph.js
@@ -35,6 +35,13 @@ module.exports = {
             taskName: 'Task.Catalog.GenerateSku'
         },
         {
+            label: 'generate-mgmt-tag',
+            waitOn: {
+                'discovery-mgmt-graph': 'succeeded'
+            },
+            taskName: 'Task.Catalog.GenerateTag'
+        },
+        {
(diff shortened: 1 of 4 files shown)


heckj:
question - is it worth adding this task into the switch or pdu discovery graphs as well as the compute ones?

zyoung51:
@heckj That's a great question.  I do not have an answer or opinion on it.  
The tags do attempt to apply to all nodes, regardless of type, but the rules should restrict it to a node type by the catalog name.  I'll kick off a thread in google groups on it?

heckj:
yeah, sounds good. Seems like it would be good to keep it consistent across all the kinds of nodes we support (`compute`, `switch`, & `pdu`) but the only concrete use case we have today is on `compute` - the largest of the herds of machines we'll manage

jlongever:
test this please

jlongever:
test this please

jlongever:
test this please

jlongever:
test this please

jlongever:
+1

brianparry:
:+1:

PR #66, opened 2016-02-26
Created a graph that sets up the BMC credentials

tannoa2:
One task is to run a template that creates a BMC username and password. The second task updates the mongoDB with new obmSettings.

### DIFF
--- a/lib/graphs/discovery-graph.js
+++ b/lib/graphs/discovery-graph.js
@@ -78,31 +78,17 @@ module.exports = {
             ignoreFailure: true
         },
         {
-            label: 'set-bmc',
-            taskName: 'Task.Set.BMC.Credentials',
-            waitOn: {
-                'catalog-lldp': 'finished'
-            },
-        },
-        {
-            label: 'create-ipmi-obm-settings',
-            taskName: 'Task.Obm.Ipmi.CreateSettings',
-            waitOn: {
-                'set-bmc': 'finished'
-            }
-        },
-        {
             label: 'shell-reboot',
             taskName: 'Task.ProcShellReboot',
             waitOn: {
-                'create-ipmi-obm-settings': 'finished'
+                'catalog-lldp': 'finished'
             }
         },
         {
             label: 'finish-bootstrap-trigger',
             taskName: 'Task.Trigger.Send.Finish',
             waitOn: {
-                'shell-reboot': 'finished'
+                'catalog-lldp': 'finished'
             }
         }
     ]
--- a/lib/graphs/discovery-graph.js
+++ b/lib/graphs/discovery-graph.js
@@ -78,17 +78,31 @@ module.exports = {
             ignoreFailure: true
         },
         {
+            label: 'set-bmc',
+            taskName: 'Task.Set.BMC.Credentials',
+            waitOn: {
+                'catalog-lldp': 'finished'
+            },
+        },
+        {
+            label: 'create-ipmi-obm-settings',
+            taskName: 'Task.Obm.Ipmi.CreateSettings',
+            waitOn: {
+                'set-bmc': 'finished'
+            }
+        },
+        {
             label: 'shell-reboot',
             taskName: 'Task.ProcShellReboot',
             waitOn: {
-                'catalog-lldp': 'finished'
+                'create-ipmi-obm-settings': 'finished'
             }
         },
         {
             label: 'finish-bootstrap-trigger',
             taskName: 'Task.Trigger.Send.Finish',
             waitOn: {
-                'catalog-lldp': 'finished'
+                'shell-reboot': 'finished'
             }
         }
     ]
--- a/lib/graphs/set-bmc-credentials-graph.js
+++ b/lib/graphs/set-bmc-credentials-graph.js
@@ -1,0 +1,68 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Set BMC Credentials',
+    injectableName: 'Graph.Set.Bmc.Credentials',
+    options: {
+        'generate-pass': {
+            user: null,
+            password: null
+        }
+    },
+    tasks: [
+        {
+            label: 'generate-pass',
+            taskDefinition: {
+                friendlyName: 'Generate BMC Password',
+                injectableName: 'Task.Generate.BMC.Password',
+                implementsTask: 'Task.Base.Generate.Password',
+                properties: { },
+                options: {
+                    user: null
+
+                }
+            }
+        },
+        {
+            label: 'set-bmc',
+            taskName: 'Task.Set.BMC.Credentials',
+            waitOn: {
+                'generate-pass': 'succeeded'
+            },
+            ignoreFailure: true
+        },
+
+        {
+            label: 'catalog-bmc',
+            taskName: 'Task.Catalog.bmc',
+            waitOn: {
+                'set-bmc': 'succeeded'
+            },
+            ignoreFailure: true
+        },
+        {
+            label: 'create-ipmi-obm-settings',
+            taskName: 'Task.Obm.Ipmi.CreateSettings',
+            waitOn: {
+                'catalog-bmc': 'succeeded'
+            }
+        },
+        {
+            label: 'shell-reboot',
+            taskName: 'Task.ProcShellReboot',
+            waitOn: {
+                'create-ipmi-obm-settings': 'finished'
+            }
+        },
+
+        {
+            label: 'finish-bootstrap-trigger',
+            taskName: 'Task.Trigger.Send.Finish',
+            waitOn: {
+                'create-ipmi-obm-settings': 'finished'
+            }
+        }
+    ]
+};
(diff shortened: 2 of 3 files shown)


tannoa2:
@RackHD/corecommitters @uppalk1 @keedya @zyoung51 @jlongever

lib/graphs/discovery-graph.js
```diff
@@ -78,18 +78,32 @@ module.exports = {
             ignoreFailure: true
         },
         {
+            label: 'set-bmc',
```
jlongever: put this before catalog-bmc

keedya:
:+1:

jlongever:
test this please

tannoa2:
Hi @jlongever , this PR will keep failing until the other two PRs are merged, becuase the graph is look for two new tasks

benbp:
@tannoa2 when you have dependent PRs will you add links to them in the PR description please?

jlongever:
@tannoa2 got you,  and what @bbp said.   Thanks,  i think this is a +1 for me after the set-bmc task is moved before the bmc catalog task.

lib/graphs/discovery-graph.js
tannoa2: @jlongever  , when I move the set-bmc up in the workflow (before catolog-bmc), I am seeing an intermittent failure with the rendering of the ‘options’ of the set-bmc task.
In the failing scenario, options data of ‘set-bmc’ is getting replaced by that of ‘Task.Linux.Bootstrap.Ubuntu’ ,as show in the json below. The option data should be user and password instead

I believe it has to do with the fact that ‘Task.Linux.Bootstrap.Ubuntu’ is not done yet.

I am not sure how to fix that. @jlongever & @benbp  any thoughts ?
 { 

kernelFile: 'vmlinuz-3.13.0-32-generic',
initrdFile: 'initrd.img-3.13.0-32-generic',
 kernelUri: 'http://172.31.128.1:8080/common/vmlinuz-3.13.0-32-generic',
initrdUri: 'http://172.31.128.1:8080/common/initrd.img-3.13.0-32-generic',18:15:23 http.1   |    basefs: 'common/base.trusty.3.13.0-32-generic.squashfs.img',
overlayfs: 'common/discovery.overlay.cpio.gz',
profile: 'linux.ipxe',
comport: 'ttyS0',
 triggerGroup: 'bootstrap' 
}

benbp:
I agree this is needed functionality but I'm not sure how I feel about adding backdoor IPMI credentials _by default_ to every discovery run. Also if it's run via discovery it means users have no control over the user/password because we hardcode the options for the discovery graph when we run it in on-http. I think this would be much better as an optional add-on step that's off by default. 

Perhaps something like:

- we have a default OBM credentials field in the config.json
- If they aren't defined, we don't ever set credentials
- If they are defined, then we do set what the user defined.

I think this is a -1 for me until this is talked about a little more.

tannoa2:
@jlongever @benbp  My bad for not mentioning the dependency, totally skipped my mind.

jlongever:
@benbp I think putting it in the discovery graph would allow the default user to get cataloged, and perhaps be part of a SKU definition.  But maybe we should start this off with a separate workflow to at least allow an admin to configure/reset a BMC user.

benbp:
Yeah, at the very least I think having it as a separate workflow that can be included within the SKU discovery graph (like alongside the nested create ipmi pollers graph) and also run on its own would accommodate a larger variety of use cases right off the bat.

benbp:
If we're taking SKUs into account, then we could also take advantage of some of the ideas around task option rendering based on environment variables (config, sku, etc.)

keedya:
Agreed with @benbp. @tannoa2 you probably can use the same OBM credentials for set BMC credentials task

tannoa2:
It seems that the option of getting the credentials  from the config.json gives the user control over setting the username and password and at the same time meet the goal of this story and is which to automatically set the credentials.

heckj:
Philosophically, I think this is likely better as it's own workflow (perhaps invoked by a SKU definition follow-on) rather than as a default workflow. Primarily because the default workflow is intentionally made to be passive and collect information, and this would be explicitly adding very active pieces to it.

The follow-on of "invoke and do XYZ" was set up so that when we identified a relevant piece of hardware, we could react to it - so this kind of thing in an example SKUpack workflow that was invoked when we recognized a specific SKU would make sense as well.

I think this is good stuff, just best in just it's own workflow that can be invoked separately, or re-invoked at will (for the use case of "we want to change all our BMC user/password credentials" - wether that's Redfish or IPMI in the future).

tannoa2:
It seems to me that having a separate workflow to set the bmc credentials outside the discovery defeat the purpose of what we are trying to achieve and which is making it easier to the user. 
A separate workflow will require the user to manually reset the node to let the microkernel boot again, it just seems little cumbersome.
What about the initial suggestion from Ben, getting the credentials from the config.json ?

zyoung51:
The credentials in config.json need to be encrypted, salted, hashed, or something'ed, they cannot be plaintext. 

I still prefer an empty default password and us auto-generating random passwords that we then sync into the database and manage independent of the user. Doing it that way means only rackhd knows the password and it is unique per node

tannoa2:
Hi @benbp, would Zach's suggestion above of having an auto generated password address your security concerns of having a default password. If so, we will go ahead and implement it.

@heckj , I agree with you that this step of adding an IPMI username/password is somewhat  active, but I still believe that it belongs in the discovery graph, since it is a crucial step to anything the user would want to do after discover . 

I want to reiterate that a separate workflow dedicated to create an IPMI user would require a manual reboot of the node since there are no OBM settings as of yet that could be used by the system. This step just seems a significant inconvenience since it requires the user to run a separate workflow and rebooting the node.  

I personally agree with Zach's suggestion since it addresses the security concerns while maintaining the spirit of this story and which is making this step easier/seamless to the user.

Any objections?

benbp:
I'm not as concerned with security as I am about making permanent, non-obvious changes to a user's system without their explicit consent/awareness. I think it's fine to add it to discovery as long as it can be enabled and disabled at the configuration level. It's a great convenience to add these credentials, but I think it has to be a opt-in, disabled by default. The majority of the users of our RackHD today are using it with systems that already have pre-defined OBM credentials, because they contain legacy infrastructure and don't have the luxury/desire to start completely from scratch.

As I think more about this, it might make sense to move the shell reboot out of discovery and into the discovery-sku graph, that way we can keep the OBM credential setting task separate from the discovery-graph itself, which I think makes more sense to keep as a passive graph in itself.

Also, if we add it to discovery, then we *have* to make it idempotent in the on-http PR (forgive me if you've already done this). Otherwise, all of our test nodes are going to quickly fill up their IPMI user list. On the topic of randomly generated passwords, does IPMI have the capability to change the password for existing users or do you have to create a new user each time (I remember this being a thorny area, but I could be wrong). If so, it seems like we'd want to blow away any existing monorail users if we're changing to a new random password (RackHD refresh, etc.).

amymullins:
@benbp and @heckj   The use case this is satisfying came about during MFG test when virgin hardware is initially configured.   The win would be in setting up the system so this doesn’t have to be done manually and not require additional reboots.  Additional reboots can be expensive, this was always an area we always honed in on when exploring possible test time reductions.  This did come up during the MFG effort, but think datacenters could benefit from this when new hardware is introduced to the configuration. 

A separate workflow will be created (that was one of the goals to be satisfied) but having that separate workflow still does not resolve the original issue with requiring manual configuration of these settings plus reboots.  There is no way to send a command to issue the reboot if running a separate workflow.

In your feedback on “ permanent, non-obvious changes”, the user will have the capability to change and/or delete the username and password.  Adding this support in to discovery by default allows for discovery to complete faster with a less manual process on top of users having the capability to manage these settings (if they need to).

heckj:
If this is set up and configured in a sky pack definition, so run that sense requires some user configuration, so that default actions that change a machine are obvious, I'm fine with it.

I would even be fine with an alternate default discovery workflow that could be placed into config.json for the overall system, but having a out of the box default that changes machines as it interrogates then is something I'm against. I don't want to disallow a user from setting up a system that will do that, but it's not an appropriate project default.

Sent from Outlook Mobile<https://aka.ms/qtex0l>




On Tue, Mar 1, 2016 at 7:13 AM -0800, "amymullins" <notifications@github.com<mailto:notifications@github.com>> wrote:


@benbp<https://github.com/benbp> and @heckj<https://github.com/heckj> The use case this is satisfying came about during MFG test when virgin hardware is initially configured. The win would be in setting up the system so this doesn’t have to be done manually and not require additional reboots. Additional reboots can be expensive, this was always an area we always honed in on when exploring possible test time reductions. This did come up during the MFG effort, but think datacenters could benefit from this when new hardware is introduced to the configuration.

A separate workflow will be created (that was one of the goals to be satisfied) but having that separate workflow still does not resolve the original issue with requiring manual configuration of these settings plus reboots. There is no way to send a command to issue the reboot if running a separate workflow.

In your feedback on “ permanent, non-obvious changes”, the user will have the capability to change and/or delete the username and password. Adding this support in to discovery by default allows for discovery to complete faster with a less manual process on top of users having the capability to manage these settings (if they need to).

jlongever:
+1

tannoa2:
this this please

tannoa2:
test this please

jlongever:
+1

PR #67, opened 2016-02-26
Rename generate-tag graph to match glob load pattern

benbp:
Glob loading looks for `*-graph.js` otherwise this graph won't get loaded.

@zyoung51 @RackHD/corecommitters


heckj:
:+1: 
Glob loading strikes again!

Maybe we should add a stanza about that in our developer guide? 

(you guys are going to think I'm all docs-crazy, but I'm trying to figure out how to expose these development idioms and conventions in a reasonable way)

zyoung51:
+1

benbp:
@heckj or maybe just in the on-taskgraph README or even a README within lib/graphs that says "if you want to add a graph name it like this." In general I think keeping smaller documentation details within the repos like this is probably better, otherwise wouldn't we have a low signal to noise ratio in the readthedocs guide?

heckj:
I'd be fine with a README addition, as long as we're consistent. I'd keep it top level, already seeing a pattern of people not looking for README's in subdirectories in https://github.com/rackhd/rackhd. Top level yeah, but stuff below seems to get quickly missed

benbp:
Ok I'll add that to this PR.

zyoung51:
@benbp a comment or logger.info about the globbing here might be cool too: https://github.com/RackHD/on-taskgraph/blob/master/lib/loader.js#L71

anhou:
+1

benbp:
Makes sense @zyoung51 I'll also add that in to this PR.

benbp:
@heckj @zyoung51 updated for your comments. Added a README entry, JSDoc comments to the loader file, and updated the logger statements.

heckj:
Thanks, :+1:

PR #77, opened 2016-03-15
Disable isc dhcp lease poller

anhou:
Disable isc dhcp lease poller for default, because 
@tldavies 's PR https://github.com/RackHD/on-dhcp-proxy/pull/28 for settting dhcp ip address lookups and @jlongever 's PR for set static ip address lookups are all merged, I think the isc dhcp poller service could be disabled now, 

Considering about running smoothly after these two merges just recently and avoid possible bugs, still leave these two file there https://github.com/RackHD/on-tasks/blob/master/lib/jobs/isc-dhcp-lease-poller.js  , https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/isc-dhcp-poller-service-graph.js. and http://rackhd.readthedocs.org/en/latest/rackhd/index.html#static-dhcp-configuration  's disable service part is  not updated temporally. if RackHD runs  no problem after several days, let's remove these two file and update docs. if anyone think it's better do it now in this PR, let me know.

@RackHD/corecommitters @tldavies @jlongever

### DIFF
--- a/lib/graphs/isc-dhcp-poller-service-graph.js
+++ b/lib/graphs/isc-dhcp-poller-service-graph.js
@@ -5,7 +5,7 @@
 module.exports = {
     friendlyName: 'isc-dhcp leases poller',
     injectableName: 'Graph.Service.IscDhcpLeasePoller',
-    serviceGraph: true,
+    serviceGraph: false,
     options: {
         'isc-dhcp-lease-poller': {
             schedulerOverrides: {


benbp:
We still rely on the leases to get BMC addresses handed out by DHCP (i.e. if they change after cataloging), and we'll also rely on it for the forthcoming switch discovery work that @pscharla is working on. That said, @tldavies is moving this code into on-dhcp-proxy anyways and I expect would delete this graph from here as part of that work?

anhou:
close it according to @benbp 's comments

PR #82, opened 2016-04-01
modify redfish discovery to optionally catalog EMC elements

zyoung51:
- Modify the redfish discovery graph to also catalog EMC
  elements on freshly added chassis/systems when auto
  catalog is enabled

### DIFF
--- a/lib/graphs/redfish-discovery-graph.js
+++ b/lib/graphs/redfish-discovery-graph.js
@@ -8,12 +8,42 @@ module.exports = {
     options: {
         defaults: {
             uri: null
+        },
+        'when-catalog-emc' : {
+            autoCatalogEmc: 'false',
+            when: '{{options.autoCatalogEmc}}',
         }
     },
     tasks: [
         {
+            'x-description': 'Enumerate the redfish endpoint',
             label: 'redfish-client-discovery',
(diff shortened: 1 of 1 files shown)


zyoung51:
@benbp Have you seen the heartbeat failure in task-runner before: https://travis-ci.org/RackHD/on-taskgraph/jobs/120125182 ?

benbp:
@VulpesArtificem see the comment above ^^. @zyoung51 I bet it's an error with using timers in the test environment.

benbp:
Although I would have thought the `.take(5)` call at https://github.com/RackHD/on-taskgraph/blob/master/spec/lib/task-runner-spec.js#L252 would completely remove the possibility of calling the stub 6 times??

lib/graphs/redfish-discovery-graph.js
```diff
@@ -8,12 +8,42 @@ module.exports = {
     options: {
         defaults: {
             uri: null
+        },
+        'when-catalog-emc' : {
+            autoCatalogEmc: 'false',
```
yyscamper: The `Boolean` type should be better than `String`

yyscamper:
+1

zyoung51:
test this please

tannoa2:
+1

PR #85, opened 2016-04-11
Feature/use local package json

davidzuhn:
This change copies the local package.json file to the packagebuild directory, and does so because that file might be modified during the manifest build to reference dependency modules that are satisfied using references from the manifest file.

For example, the dependency of
```
git+https://github.com/RackHD/on-core.git
```
might be changed to:
```
on-core": "/emc/zuhnd/test-manifest-build/builddir/onrack-test/on-core"
```
during the build process, and then npm will use the version of on-core that's been checked out as per the detailed build manifest.

This change also ensures that errors in the build process are propagated upward (via set -e) so that failures in the ./HWIMO-BUILD process will be noticed upstream.

### DIFF
--- a/HWIMO-BUILD
+++ b/HWIMO-BUILD
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2015, EMC, Inc.
+# Copyright 2015, 2016, EMC, Inc.
 
 # debian packages expected...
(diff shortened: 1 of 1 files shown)


benbp:
👍

benbp:
test this please

benbp:
@RackHD/corecommitters

PR #86, opened 2016-04-15
Overwrite kernel and initrd file to 3.13.0 when flashing quanta

iceiilin:
"catalog ami" task before flashing Quanta BIOS would fail in command "./afulnx_64" in microkernel complaining no kernel source exist, thus the following "flash BIOS" task would not be executed. Overwriting the "bootstrap ubuntu" option to 3.13.0 version solve this problem.

Change to use megaraid catalog instead of ami for post-megaraid-update-catalog. Change to use bmc catalog instead of ami for post-bmc-update-catalog.

@benbp

### DIFF
--- a/lib/graphs/flash-quanta-all-graph.js
+++ b/lib/graphs/flash-quanta-all-graph.js
@@ -11,7 +11,9 @@ module.exports = {
         },
         'bootstrap-ubuntu': {
             basefs: "common/base.trusty.3.13.0-32.squashfs.img",
-            overlayfs: 'common/overlayfs_quanta_t41_flash-v1.1-3.13.0-32.cpio.gz'
+            overlayfs: 'common/overlayfs_quanta_t41_flash-v1.1-3.13.0-32.cpio.gz',
+            kernelFile: 'vmlinuz-3.13.0-32-generic',
+            initrdFile: 'initrd.img-3.13.0-32-generic'
         },
         'download-megaraid-firmware': {
             file: null
@@ -130,7 +132,7 @@ module.exports = {
         // Post-catalog
         {
             label: 'catalog-quanta-megaraid-after',
-            taskName: 'Task.Catalog.ami',
+            taskName: 'Task.Catalog.megaraid',
             waitOn: {
                 'flash-bmc': 'succeeded'
             }
@@ -144,7 +146,7 @@ module.exports = {
         },
         {
             label: 'catalog-quanta-bmc-after',
-            taskName: 'Task.Catalog.ami',
+            taskName: 'Task.Catalog.bmc',
             waitOn: {
                 'catalog-quanta-bios-after': 'succeeded'
             }
--- a/lib/graphs/flash-quanta-bmc-graph.js
+++ b/lib/graphs/flash-quanta-bmc-graph.js
@@ -63,7 +63,7 @@ module.exports = {
         },
         {
             "label": "catalog-quanta-bmc-after",
-            "taskName": "Task.Catalog.ami",
+            "taskName": "Task.Catalog.bmc",
             "waitOn": {
                 "flash-bmc": "succeeded"
             }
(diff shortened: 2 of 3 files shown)


iceiilin:
test this please.

iceiilin:
travis CI failure is in task-runner-spec about heart beat interval in node 4.

stuart-stanley:
-1
I believe the there is a push to move towards using afulnx instead of storcli to manage these. My guess is that afulnx will be in the kernel bump that should come in today, and this problem can then be fixed by just referencing the new kernel (if the kernel bump PR's don't already modify this file).

benbp:
@stuart-stanley not quite sure what you mean? My understanding is that afulnx and storcli are completely separate tools, one for managing Intel BIOS settings, the other for managing LSI RAID devices. Are you also saying that afulnx ships in the 3.19 kernel by default? The reason we have the 3.13 kernel is because the flash workflows still use an old, pre-built image that has the afu modules and tools baked into it.

zyoung51:
The afulnx binary in the new Quanta BIOS packages does not have the same kernel source requirement.  Overlayfs is no longer used here: https://github.com/RackHD/on-skupack/blob/master/quanta-t41/workflows/flash-quanta-bios-graph.json.

lib/graphs/flash-quanta-all-graph.js
```diff
@@ -130,7 +132,7 @@ module.exports = {
         // Post-catalog
         {
             label: 'catalog-quanta-megaraid-after',
-            taskName: 'Task.Catalog.ami',
+            taskName: 'Task.Catalog.megaraid',
```
stuart-stanley: (putting context on my main comments)
this change here moves from using ami (which uses aflinux) to using megaraid.

lib/graphs/flash-quanta-all-graph.js
```diff
@@ -144,7 +146,7 @@ module.exports = {
         },
         {
             label: 'catalog-quanta-bmc-after',
-            taskName: 'Task.Catalog.ami',
+            taskName: 'Task.Catalog.bmc',
```
stuart-stanley: this changes catalogs the bmc rather than doing the megaraid via ami twice. I.E. it fixes an "oops" not mentioned in the PR description.

lib/graphs/flash-quanta-bmc-graph.js
```diff
@@ -63,7 +63,7 @@ module.exports = {
         },
         {
             "label": "catalog-quanta-bmc-after",
-            "taskName": "Task.Catalog.ami",
+            "taskName": "Task.Catalog.bmc",
```
stuart-stanley: again, an unmentioned "oops" fix.

stuart-stanley:
@benbp and @zyoung51 added comments about the changes to maybe help set context.

Right now there appears to be a split between functionality in skupacks vs "in place". hrm. ack. I don't have time to untangle this right now though. I *will* leave the 3.13 files in place in case this is needed for now.

iceiilin:
@stuart-stanley @zyoung51 This PR is a quick fix for OnRack 1.2.0 release. I didn't find a 1.2.0 release branch so I opened it  against the master.

I didn't notice the FW update feature in skupack before, and it might bring some confusion when reviewing this PR. Currently OnRack still use the old graph I modified here instead of the skupack. As @benbp said, the pre-build microkernel for this graph has the afu modules based on 3.13.0. 

I checked the FW update in skupack, fetching this tools runtime when flashing without relying on microkernel is certainly more convenient and flexible, but I don't think we need to apply this mechanism  in this graph that will gradually be deprecated.

lib/graphs/flash-quanta-all-graph.js
iceiilin: It is a post catalog for megaraid update in "flash all" graph, so megaraid catalog will be used here to verify the new FW version.

lib/graphs/flash-quanta-all-graph.js
iceiilin: Will add in description

stuart-stanley:
yeah. +1

yyscamper:
:+1:

anhou:
:+1:

PR #87, opened 2016-04-20
add default 'repo' in install esxi graph

yyscamper:
Add default `repo` for install ESXi graph to align with other OSs, such as [RHEL](https://www.github.com/RackHD/on-taskgraph/blob/master/lib/graphs/install-rhel-graph.js#L11), [CentOS](https://www.github.com/RackHD/on-taskgraph/blob/master/lib/graphs/install-centos-graph.js#L11), [SUSE](https://www.github.com/RackHD/on-taskgraph/blob/master/lib/graphs/install-suse-graph.js#L11).

The benefit we get is that if user follows our default address to setup the OS repo, then only the `version` is required for OS bootstrap, such make the OS bootstrap pretty simple.

@RackHD/corecommitters @pengz1 @iceiilin @WangWinson @cgx027 @sunnyqianzhang

### DIFF
--- a/lib/graphs/install-esx-graph.js
+++ b/lib/graphs/install-esx-graph.js
@@ -8,7 +8,7 @@ module.exports = {
     options: {
         defaults: {
             version: null,
-            repo: null
+            repo: '{{api.server}}/esxi/{{options.version}}'
         },
         'install-os': {
             schedulerOverrides: {


anhou:
:+1:

WangWinson:
@yyscamper should we add default `version` value, too ?

yyscamper:
@WangWinson : I  intended to not add the default `version` because I think it is not reasonable and we don't have any preference on which `version` is better.

WangWinson:
@yyscamper my concern, if there is no default `version` there, the install OS workflow will fail directly. I have find that when I try some OS install workflow without any `-d @param.json`

benbp:
@WangWinson that's the intent of the null defaults, because I think more often than not a user will specify the version, given that they have to download/mirror all the files for that version on their own anyways.

benbp:
Oh, looks like you +1ed it anyways. Merging :)

PR #88, opened 2016-04-20
 Overwrite kernel and initrd file to 3.13.0 in quanta bmc update workflow

iceiilin:
Overwrite kernel and initrd version to 3.13.0. The tool (soc_flash) for updating BMC is not relied on kernel, while the ipmitool used for catalog before and after updating uses ipmi driver in kernel. 

Remove reboot node task at the end of this workflow. The operation is not a requirement for updating BMC, as suggested in the updating script provided by Quanta. Meanwhile, after updating, BMC network configuration is reset to default (DHCP), so this operation would fail in some network environment, if its IP is not assigned by DHCP server in RackHD control network. Per @zyoung51, there would be a story about saving the configuration before updating and applying it to the new BMC.

### DIFF
--- a/lib/graphs/flash-quanta-bmc-graph.js
+++ b/lib/graphs/flash-quanta-bmc-graph.js
@@ -11,7 +11,9 @@ module.exports = {
         },
         "bootstrap-ubuntu": {
             "basefs": "common/base.trusty.3.13.0-32.squashfs.img",
-            "overlayfs": "common/overlayfs_quanta_t41_flash-v1.1-3.13.0-32.cpio.gz"
+            "overlayfs": "common/overlayfs_quanta_t41_flash-v1.1-3.13.0-32.cpio.gz",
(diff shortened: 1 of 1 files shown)


yyscamper:
:+1:

benbp:
@iceiilin I think the intent of the final reboot was so the node didn't hang in the microkernel indefinitely?

iceiilin:
@benbp I understand the intention of the reboot task in flashing BMC, but since BMC network configuration resets to default (DHCP) after flashing, this would fail in some hardware environment (static BMC IP) and the system would remain in microkernel. So I would prefer to remove it and make the workflow success. We can add the reboot task after @zyoung51 's story to preserve BMC configuration during flashing.
What's your idea? Would you like to reboot the system with "DHCP" BMC and leave the workflow failing in "static" BMC?

anhou:
@iceiilin assume you've synced with @zyoung51  and @benbp and reach an agreement  :+1:

PR #90, opened 2016-04-25
Add validate ssh task to OS install graphs

benbp:
@RackHD/corecommitters @heckj @VulpesArtificem

### DIFF
--- a/lib/graphs/install-centos-graph.js
+++ b/lib/graphs/install-centos-graph.js
@@ -1,4 +1,4 @@
-// Copyright 2015, EMC, Inc.
+// Copyright 2015-2016, EMC, Inc.
 
 'use strict';
 
@@ -14,6 +14,9 @@ module.exports = {
             schedulerOverrides: {
                 timeout: 3600000 //1 hour
             }
+        },
+        "validate-ssh": {
+            retries: 10
         }
     },
     tasks: [
@@ -34,6 +37,13 @@ module.exports = {
             taskName: 'Task.Os.Install.CentOS',
             waitOn: {
                 'reboot': 'succeeded'
+            }
+        },
+        {
+            label: "validate-ssh",
+            taskName: "Task.Ssh.Validation",
+            waitOn: {
+                "install-os": "succeeded"
             }
         }
     ]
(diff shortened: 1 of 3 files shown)


heckj:
👍

VulpesArtificem:
👍

benbp:
test this please

lib/graphs/install-centos-graph.js
```diff
@@ -35,6 +38,13 @@ module.exports = {
             waitOn: {
                 'reboot': 'succeeded'
             }
+        },
+        {
+            label: "validate-ssh",
+            taskName: "Task.Ssh.Validation",
```
yyscamper: I think the RackHD appliance may not always have the authority to login into the target OS, the targetOS may be configured in another network with static IP, which is disconnect with RackHD, so how about add a `ignoreFailure` to this task?

yyscamper:
:+1: after my comment is fixed or clarified.

lib/graphs/install-centos-graph.js
benbp: @yyscamper I think in those cases, the graph could also be configured to pass in an empty users value to the validate-ssh task, or define a new graph that does not do the validation (or not have the validation included in the default install-centos graph).

I'm hesitant to add ignoreFailure, because then we have no way of catching other failures that should not be ignored (like failure to connect with settings we expect to work).

anhou:
:+1:

lib/graphs/install-centos-graph.js
yyscamper: @benbp : I agree the `ignoreFailure` will not be a good solution as it will ignore all kinds of errors.

By checking the code, I don't find there is a possibility that user can pass an empty user  to skip the validate-ssh task, because we used `concat` in both places:
- https://github.com/RackHD/on-tasks/blob/master/lib/jobs/validate-ssh.js#L52
- https://github.com/RackHD/on-tasks/blob/master/lib/jobs/install-os.js#L110 

Since the root account is required for install-os task, so this validation task will always get an account to do validation, but this will cause whole install-os graph fails due to the network problem I mentioned above.

What's more, it surely works to create another workflow, but I don't think it is a good one because of the duplication.

So how about to update the ssh-validation job to accept a flag to skip the validation?

benbp: @yyscamper the _.isEmpty check at https://github.com/RackHD/on-tasks/pull/173/files#diff-8ee10fe8a2f7f28a0644338429d6b4f3R58 should handle an empty users array. Optionally, one could use a separate options object to specify users for the centos task rather than "defaults".

PR #91, opened 2016-04-28
Arista specific graphs for switch discovery and ZTP

pscharla:
New graphs for Arista catalog and deploy of startup-config script

@RackHD/corecommitters @benbp @heckj @VulpesArtificem @zyoung51


lib/graphs/deploy-arista-switch-startup-config-graph.js
```diff
@@ -0,0 +1,33 @@
...
+            "taskDefinition": {
+                "friendlyName": "Deploy switch startup config",
+                "injectableName": "Task.Inline.Switch.Deploy.StartupConfig",
+                "implementsTask": "Task.Base.Linux.Commands",
+                "options": {
+                    "startupConfig": null,
+                    "startupConfigUri": "{{ api.base }}/templates/{{ options.startupConfig }}",
+                    "commands": [
+                        {
+                            "downloadUrl": "/api/1.1/templates/arista-deploy-startup-config.py"
```
benbp: Should you be using the `{{ startupConfigUri }}` value here instead?

benbp: Actually, does this need to be included given that you have an example here: https://github.com/RackHD/on-skupack/pull/14/files ?

pscharla: I'm not 100% sure. I noticed that the sku pack workflows had an equivalent graph in on-taskgraph, so that is why I put it there. But maybe not needed since it is deploying the startup config file (probably a one time thing) versus what I was looking at specifically was workflows that get BIOS and update BIOS on a compute node (could be done multiple times).

pscharla: @zyoung51 Are workflows included in the skupacks typically placed in on-taskgraph as well? I noticed that pattern looking at other SKUs, but not sure if it's necessary.

zyoung51: @pscharla Nope, the sku specific workflows should only go in the sku pack and not in on-taskgraph at all.  The Quanta workflows originated in on-taskgraph and were moved into sku packs, but have not been removed from on-taskgraph yet.

pscharla: @zyoung51 OK. Thanks. Will remove here then.

benbp:
👍

zyoung51:
This should be javascript instead of json?  @benbp should we add .json to the file globbing or use the file loaders instead of require to differentiate between the two?

benbp:
@zyoung51 it will work either way. We also support just loading .json files.

yyscamper:
:+1:

PR #92, opened 2016-04-29
Bump mocha timeout to 10000 to mitigate potential jenkins issues

benbp:
Proposing a potential way to address what could be jenkins timeout issues when loading modules in mocha "before" blocks.

@RackHD/corecommitters @tldavies @davidzuhn

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "lib",
   "scripts": {
     "doc": "jsdoc lib -r -d docs",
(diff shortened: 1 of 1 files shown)


tldavies:
👎 This should try to be fixed via jenkins with lowering the node count for the build slave first. As it is not breaking on RackHD. So I think this is the build machines fault. More comments on RackHD/on-core#128

PR #95, opened 2016-05-05
Add ARP Poller Graph

jlongever:
@RackHD/corecommitters @uppalk1 @zyoung51 @tannoa2 @keedya @geoff-reid 

Dependent PR https://github.com/RackHD/on-tasks/pull/182

### DIFF
--- a/lib/graphs/arp-poller-service-graph.js
+++ b/lib/graphs/arp-poller-service-graph.js
@@ -1,0 +1,24 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'ARP poller',
+    injectableName: 'Graph.Service.ArpPoller',
+    serviceGraph: true,
+    options: {
+        'arp-poller': {
+            schedulerOverrides: {
+                timeout: -1
+            }
+        }
+    },
+    tasks: [
+        {
+            label: 'arp-poller',
+            taskName: 'Task.ArpPoller'
+        }
+    ]
+};
+
+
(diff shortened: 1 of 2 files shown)


brianparry:
test this please

lib/graphs/arp-poller-service-graph.js
```diff
@@ -0,0 +1,24 @@
...
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'ARP poller',
+    injectableName: 'Graph.Service.ArpPoller',
+    serviceGraph: false,
+    options: {
+        'arp-poller': {
+            schedulerOverrides: {
```
benbp: FYI you can also just do `'arp-poller': { timeout: -1 }` now.

lib/graphs/arp-poller-service-graph.js
```diff
@@ -0,0 +1,24 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'ARP poller',
+    injectableName: 'Graph.Service.ArpPoller',
+    serviceGraph: false,
```
benbp: I assume your intent is to allow a user to enable it if they want, rather than just always running it?

benbp:
👍

lib/graphs/arp-poller-service-graph.js
jlongever: @benbp Yeah,  having both DHCP poller and ARP poller running a the same time should be fine though. You'll probably see more lookup activity with the ARP poller since the entries are updated more frequently than the DHCP lease cache.

jlongever: Should I just set this to true?  It was true originally and then I changed it.

benbp: Works for me. @tldavies FYI I think this may actually remove the need for the dhcp lease poller entirely (runs and hides).

benbp:
test this please

yyscamper:
:+1:

jlongever:
test this please

anhou:
:+1:

PR #96, opened 2016-05-09
added unique flags to indexObject

VulpesArtificem:
depends on https://github.com/RackHD/on-core/pull/132
changed so that we only create the necessary compound indexes in task-graph-runner; other indexes and uniqueness enforcement have been moved to the waterline models.  
@RackHD/corecommitters


anhou:
:+1:

benbp:
👍

PR #99, opened 2016-05-13
Add task to centos installer that waits for an API call on reboot bef…

benbp:
…ore SSH validation

Add an extra task to the CentOS installer that waits until after the reboot stage before doing SSH validation.

Requires https://github.com/RackHD/on-http/pull/279 and https://github.com/RackHD/on-tasks/pull/201

This also depends on @zyoung51's changes to add a completionUri wait task: https://github.com/RackHD/on-tasks/pull/196

@RackHD/corecommitters @heckj @zyoung51 @VulpesArtificem @johren @stuart-stanley

### DIFF
--- a/lib/graphs/install-centos-graph.js
+++ b/lib/graphs/install-centos-graph.js
@@ -7,15 +7,23 @@ module.exports = {
     injectableName: 'Graph.InstallCentOS',
     options: {
         defaults: {
+            // Make sure this matches for both the install-os task and the
+            // rackhd-callback-uri-wait task, so put it in "defaults"
+            completionUri: 'renasar-ansible.pub',
             version: null,
             repo: '{{api.server}}/centos/{{options.version}}/os/x86_64'
         },
         'install-os': {
             schedulerOverrides: {
-                timeout: 3600000 //1 hour
+                timeout: 3600000 // 1 hour
             }
         },
-        "validate-ssh": {
+        'rackhd-callback-uri-wait': {
+            schedulerOverrides: {
+                timeout: 1200000 // 20 minutes
+            }
+        },
+        'validate-ssh': {
             retries: 10
         }
     },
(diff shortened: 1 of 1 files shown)


yyscamper:
A wonderful solution!

But I still have a concern is that: RackHD may not have the ability to access the target OS after installation completion, due to any of following reasons:
(1) User set a wrong network address
(2) User intentionally to prevent RackHD's accessibility (user may allow RackHD to touch BMC but not OS).
(3) There is a possibility that the boot-order may not just point to the install disk, so after reboot, the system will not boot into the new OS.

So whether we need some extra option to bypass these new tasks that append at the previous workflow?

Anyway, I agree this is a very good solution to get the flag that installation really takes effect.

benbp:
@yyscamper Thanks! To your questions I would respond:

For (1): that's a real failure case and we would expect the wait on completion uri task to timeout and indicate this type of failure.
For (2): In that case, I would advise the user to make their own graph that excludes the final two tasks.
For (3): Same solution as #2 above.

I think it is a better pattern to encourage users to write their own workflows to suit their specific needs, as that is the intent of the modularity of the workflow engine, rather than trying to write one single CentOS installer workflow that will work for every possible use case we can anticipate.

jlongever:
test this please

yyscamper:
@benbp , I agree, a single workflow is hard to cover all cases, But according the design of OnRack, it currently doesn't support multiple workflows to the ESXi installation.

So in RackHD level, it's surely :+1: for me.

Please @johren , @stuart-stanley confirm whether this design meets your requirement though there is additional network configuration requirement as I listed above.

johren:
@yyscamper:  This does meet our requirements.   The expectation in this workflow is that the ORA will have network connectivity after ESXi is installed and it will boot into ESXi.  I agree with @benbp that other workflows should be created if we encounter a case where we need to support (2) or (3) above.

benbp:
@yyscamper this PR isn't meant to address ESXi installs, only CentOS. I'm actively working on ESXi, though it will look very similar to this solution.

anhou:
:+1:

yyscamper:
Thanks for your feedback @johren 

@benbp  still :+1: for me.

PR #105, opened 2016-05-23
Add SSH validation to RHEL install graph

benbp:
Adds wait for reboot and SSH validation tasks to the RHEL installer graph (identical process to the CentOS installer updates made last week).

@RackHD/corecommitters @heckj @johren @richav1

### DIFF
--- a/lib/graphs/install-rhel-graph.js
+++ b/lib/graphs/install-rhel-graph.js
@@ -7,6 +7,9 @@ module.exports = {
     injectableName: 'Graph.InstallRHEL',
     options: {
         defaults: {
+            // Make sure this matches for both the install-os task and the
+            // rackhd-callback-uri-wait task, so put it in "defaults"
+            completionUri: 'renasar-ansible.pub',
             version: null,
             repo: '{{api.server}}/rhel/{{options.version}}/os/x86_64'
         },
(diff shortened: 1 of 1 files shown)


lib/graphs/install-rhel-graph.js
```diff
@@ -7,13 +7,24 @@ module.exports = {
     injectableName: 'Graph.InstallRHEL',
     options: {
         defaults: {
+            // Make sure this matches for both the install-os task and the
+            // rackhd-callback-uri-wait task, so put it in "defaults"
+            completionUri: 'renasar-ansible.pub',
             version: null,
             repo: '{{api.server}}/rhel/{{options.version}}/os/x86_64'
```
benbp: @richav1 I noticed in my testing that the default directory structure (for rhel 7.2 at least) doesn't support this URI path (`'{{api.server}}/rhel/{{options.version}}'` is what works for me). Is this correct or do you have some different directory structure in your test environments?

heckj:
👍

lib/graphs/install-rhel-graph.js
richav1: @benbp  First of all, we only use rhel 7.0.  Secondly, yes the URI path as is works for use.

zyoung51:
+1

anhou:
+1

PR #112, opened 2016-06-01
Set service graph status to running.

rolandpoulter:
To improve consistency, since graphs created from on-http will start in a running state. Before the service graphs would be pending.


yyscamper:
:+1:

anhou:
:+1:

PR #115, opened 2016-06-06
Remove BMC user credentials graph

keedya:
- Allow the user to remove added BMC users to restore defaults
- note: This is part of decommissioning a node story. 

depends on: https://github.com/RackHD/on-tasks/pull/233
depends on: https://github.com/RackHD/on-http/pull/320

### DIFF
--- a/lib/graphs/bootstrap-bmc-credentials-remove-graph.js
+++ b/lib/graphs/bootstrap-bmc-credentials-remove-graph.js
@@ -1,0 +1,60 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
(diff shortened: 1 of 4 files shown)


keedya:
@RackHD/corecommitters @zyoung51 @jlongever @uppalk1 @tannoa2

jlongever:
Was there a reason why you didn't want to add a remove option to https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/set-bmc-credentials-graph.js using the `Task.Evaluate.Condition` task?  Example: https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/redfish-discovery-graph.js

keedya:
@jlongever  there is no reason. I think both ways are valid.

PR #118, opened 2016-06-10
Decommission graph

keedya:
@jlongever @benbp @zyoung51 

- Remove BMC users
- Allow user to select between secure erase or run  DD shell command on the node

### DIFF
--- a/lib/graphs/bootstrap-decommission-node-graph.js
+++ b/lib/graphs/bootstrap-decommission-node-graph.js
@@ -68,7 +68,7 @@ module.exports = {
             label: 'catalog-megaraid',
             taskName: 'Task.Catalog.megaraid',
             waitOn: {
-                'drive-secure-erase': 'succeeded'
+                'remove-bmc-credentials': 'finished'
             },
             ignoreFailure: true
         },
@@ -76,7 +76,7 @@ module.exports = {
             label: 'catalog-driveid',
             taskName: 'Task.Catalog.Drive.Id',
             waitOn: {
-                'catalog-megaraid': 'finished'
+                'remove-bmc-credentials': 'finished'
             },
             ignoreFailure: true
         },
@@ -93,6 +93,22 @@ module.exports = {
             waitOn: {
                 'when-secure-erase': 'succeeded'
             }
+        },
+        {
+            label: 'catalog-megaraid',
+            taskName: 'Task.Catalog.megaraid',
+            waitOn: {
+                'drive-secure-erase': 'finished'
+            },
+            ignoreFailure: true
+        },
+        {
+            label: 'catalog-driveid',
+            taskName: 'Task.Catalog.Drive.Id',
+            waitOn: {
+                'drive-secure-erase': 'finished'
+            },
+            ignoreFailure: true
         },
         {
             label: 'reboot-after-shell-commands',
--- a/lib/graphs/bootstrap-decommission-node-graph.js
+++ b/lib/graphs/bootstrap-decommission-node-graph.js
@@ -71,28 +71,28 @@ module.exports = {
         },
         {
             label: 'reboot-after-shell-commands',
-            taskName: 'Task.Trigger.Send.Finish',
+            taskName: 'Task.ProcShellReboot',
             waitOn: {
                 'shell-commands': 'finished'
             }
         },
         {
             label: 'reboot-after-drive-secure-erase',
-            taskName: 'Task.Trigger.Send.Finish',
+            taskName: 'Task.ProcShellReboot',
             waitOn: {
                 'drive-secure-erase': 'finished'
             }
         },
         {
             label: 'finish-after-shell-commands',
-            taskName: 'Task.ProcShellReboot',
+            taskName: 'Task.Trigger.Send.Finish',
             waitOn: {
                 'shell-commands': 'finished'
             }
         },
         {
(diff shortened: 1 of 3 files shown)


lib/graphs/bootstrap-decommission-node-graph.js
```diff
@@ -0,0 +1,102 @@
...
+        {
+            label: 'shell-commands',
+            taskName: 'Task.Linux.Commands',
+            waitOn: {
+                'when-secure-erase': 'failed'
+            }
+        },
+        {
+            label: 'drive-secure-erase',
+            taskName: 'Task.Drive.SecureErase',
```
yyscamper: Check the SecureErase workflow: https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/secure-erase-drive-graph.js, it actually need some a particular overlayfs and some special parameter like `eraseSettings`, meanwhile, some catalog need be refreshed after secure erase. I think the decommission workflow need do the same work. 

@pengz1 @iceiilin

lib/graphs/bootstrap-decommission-node-graph.js
```diff
@@ -0,0 +1,127 @@
...
+    injectableName: 'Graph.Bootstrap.Decommission.Node',
+    options: {
+        'shell-commands': {
+            commands:[
+                { 
+                  command: "for disk in `lsblk | grep disk | awk '{print $1}'`;\
+                    do sudo dd if=/dev/zero of=/dev/$disk bs=512 count=1 ; done"
+                }
+            ]
+        },
```
pengz1: secure erase can do all kinds of secure erase, why we implemented a new method here?
"scrub -p fillzero /dev/sdx" can implement same function

And it should be duplicated with secure erase.

lib/graphs/bootstrap-decommission-node-graph.js
```diff
@@ -0,0 +1,127 @@
...
+        'remove-bmc-credentials': {
+            users: null
+        },
+        'bootstrap-ubuntu': {
+            overlayfs: "common/secure.erase.overlay.cpio.gz",
+            triggerGroup: 'secureErase'
+        },
+        'drive-secure-erase': {
+            eraseSettings: null
```
pengz1: Use default secure erase command might be rather slow.
Default secure erase will do 3 times overwriting:
0x00, 0xff, random.
Is this expected?

yyscamper:
@keedya : I checked the functional spec, there is an additional goal for decommission, I think you have forgot this implementation.
```
all pollers for the relevant node should be stopped and removed, and active monitoring ceased
```

lib/graphs/bootstrap-decommission-node-graph.js
```diff
@@ -0,0 +1,127 @@
...
+                'when-secure-erase': 'failed'
+            }
+        },
+        {
+            label: 'drive-secure-erase',
+            taskName: 'Task.Drive.SecureErase',
+            waitOn: {
+                'when-secure-erase': 'succeeded'
+            }
+        },
```
pengz1: Does this node still exist? if so, after secure erase all drive catalogs are different from before-secure-erase (Because all RAID will be delete), you shoudl re-catalog driveId and MegaRaid info after secure-erase.

yyscamper: After decommission, the node will be removed? If not, I think you still need to refresh the catalog after decommission otherwise the node's catalog will be out-of-date.

Confirmed with @pengz1, The catalog refresh before secure erase is still required to avoid user have some manual operation on disk which not via RackHD, so when the secure erase starts, all data is update-to-date.

lib/graphs/bootstrap-decommission-node-graph.js
yyscamper: Since secure-erase is optional, so the dd operation can do a quick erase on all disks. Surely the dd is not necessary if secure-erase is enabled, but consider the dd time is very much litter than secure-erase, so always enable the dd erase seems reasonable.

lib/graphs/bootstrap-decommission-node-graph.js
keedya: @yyscamper , Since disabling pollers  already exist in an api calls we provide in RackHD, User can run it separately to disable pollers. I am open for suggestions here if you think otherwise

lib/graphs/bootstrap-decommission-node-graph.js
keedya: @pengz1, drive secure erase is optional in this workflow. If user decide to use it, he needs to use  valid and selective options to run it

lib/graphs/bootstrap-decommission-node-graph.js
yyscamper: @keedya : which API call? could you please give a link. My understanding is the decommission will be done by triggering a workflow, so all tasks should be included in a workflow.

keedya: @yyscamper , below is the api to disable a poller:
PATCH /api/common/pollers/<node Id>   body : {"paused": true}

The story is to provide user with methods via API requests to achieve decommissioning a specific node.

amymullins:
@yyscamper and @anhou , can you help confirm your questions have been answered?  I'm looking continue the  momentum of this feature and (hopefully) complete this story given tomorrow is the end of the sprint.  thanks!

PR #122, opened 2016-06-15
Modify Graph.Discovery to support sku discovery.

cgx027:
Add a task to set bootdev to pxe, so that the following run-sku-graph
will be run following a pxe boot. Please also refer to
lib/graph/discovery-sku-graph.js

### DIFF
--- a/lib/graphs/discovery-graph.js
+++ b/lib/graphs/discovery-graph.js
@@ -78,17 +78,32 @@ module.exports = {
             ignoreFailure: true
         },
         {
+            "label": "set-boot-pxe",
+            "taskDefinition": {
+                "friendlyName": "Set PXE boot",
+                "injectableName": "Task.Inline.Catalog.Switch.Arista",
+                "implementsTask": "Task.Base.Linux.Commands",
+                "options": {
+		    "commands": "sudo ipmitool chassis bootdev pxe"
+                },
+                "properties": {}
+            },
+            waitOn: {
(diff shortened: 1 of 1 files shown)


lib/graphs/discovery-graph.js
```diff
@@ -78,18 +78,33 @@ module.exports = {
             ignoreFailure: true
         },
         {
+            "label": "set-boot-pxe",
+            "taskDefinition": {
+                "friendlyName": "Set PXE boot",
+                "injectableName": "Task.Inline.Catalog.Switch.Arista",
```
yyscamper: This is an obm-control task is for set pxe boot, why use this one?

pengz1: There is no obm setting at this timeline.
However I am curious why injectableName is "Task.Inline.Catalog.Switch.Arista"?

iceiilin: The pxe boot is set in microkernel, and doesn't need obm setting.

iceiilin:
:+1: after modifying the injectableName

pengz1:
:+1: after modifying the injectableName

lib/graphs/discovery-graph.js
cgx027: @pengz1 Opps, this should be changed after copy-paste.

yyscamper:
:+1:

PR #132, opened 2016-06-25
Update HWIMO-TEST mocha timeout to 10s.

jimturnquist:
This is necessary for some build machines where dependency
loading is too slow to stay under the 2s per test timeout.

@RackHD/corecommitters

### DIFF
--- a/HWIMO-TEST
+++ b/HWIMO-TEST
@@ -14,6 +14,6 @@ npm install
 
 
 # Runs the mocha tests and reports the code coerage.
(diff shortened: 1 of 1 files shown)


benbp:
👍

benbp:
To add a note for posterity: the root cause that makes this necessary is when the source files are mounted via NFS, which sometimes introduces enough latency to make file loading during test exceed a 2 second timeout.

FYI @tldavies

tldavies:
+1

PR #134, opened 2016-06-29
tweak install-os graph's option to accommodate task-schema upfront validation

yyscamper:
When the task-schema upfront validation is enabled, all options in `defaults` will be passed into all tasks, this will cause some error in following case:
for example, if set version & repo to following format:
```
options: {
   defaults: {
     version: null,
     repo: "{{api.server}}/{{options.version}}"
   },
   install-os: { }
}
```

if user passes the workflow options by following format:
```
{
   options: {
      install-os: {
        version: 6.5,
        repo: "xxxxx"
      }
   }
}
```

The `{version: null, repo: "{{api.server}}/{{options.version}}"}` will be passed to all other tasks except install-os. But the actual `version` & `repo` is only passed to 'install-os' tasks. So for all other tasks, the task options rendering will fail as the version is `null` for {{options.version}}.

My change is to remove these kind of data from `defaults`, so the null options will not be passed into every task, then every task can success.

Depend on: https://github.com/RackHD/on-tasks/pull/257, https://github.com/RackHD/on-core/pull/171

@RackHD/corecommitters @WangWinson @iceiilin @cgx027 @pengz1

### DIFF
--- a/lib/graphs/boot-livecd-graph.js
+++ b/lib/graphs/boot-livecd-graph.js
@@ -6,13 +6,8 @@ module.exports = {
     friendlyName: 'Boot LiveCD',
     injectableName: 'Graph.BootLiveCD',
     options: {
-        defaults: {
-            repo: '{{api.server}}/LiveCD/{{options.version}}'
-        },
         'install-os': {
-            schedulerOverrides: {
-                timeout: 3600000 //1 hour
-            }
+            _taskTimeout: 3600000 //1 hour
         }
     },
     tasks: [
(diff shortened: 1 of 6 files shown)


yyscamper:
There is some unit-test error, I will fix it tomorrow.

iceiilin:
+1

yyscamper:
The unit-test error is expected, it depends on the PR: https://github.com/RackHD/on-tasks/pull/257

benbp:
👍

yyscamper:
Both Jenkins & TravisCI error is expected, because there is inter-dependency between this PR and https://github.com/RackHD/on-tasks/pull/257.

I have tested in local environment with both PRs ready, the unit-test succeed.

I will try to first merge this PR and then check the on-tasks one. If everything is OK, then I will go one with the merge of on-tasks. If fails, I will revert this on-taskgraph PR.

PR #135, opened 2016-06-29
POC Dockerized Discovery image in RACKHD

keedya:
@RackHD/corecommitters 

depends on: https://github.com/RackHD/on-http/pull/350
depends on: https://github.com/RackHD/on-tasks/pull/258

### DIFF
--- a/lib/graphs/discovery-graph.js
+++ b/lib/graphs/discovery-graph.js
@@ -6,7 +6,7 @@ module.exports = {
     friendlyName: 'Discovery',
     injectableName: 'Graph.Discovery',
     options: {
-        'bootstrap-ubuntu': {
+        'bootstrap-rancher': {
             'triggerGroup': 'bootstrap'
         },
         'finish-bootstrap-trigger': {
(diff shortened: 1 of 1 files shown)


benbp:
Before we switch over to using Rancher, I think we want/need an automated and repeatable build process for the images up and running in our repositories, and uploaded to github. Do you have changes coming in for that? Let me know if you need @heckj or I to create a new repository for you!

keedya:
@benbp , I created a separate repo to build the docker image. I think we should eventually move it to on-imagebuilder
https://github.com/keedya/rackHD-micro-docker

benbp:
@keedya whoops, yeah I forgot about that :) So my opinion is that we want that merged into RackHD and functional before doing the grand switch, that way everything is readily available for downstream consumers, and then we can start the process of migrating OEM specific images over to the container based model (which I assume you/@zyoung51 would like to see happen).

keedya:
Closing POC and openning a new PR with the final changes
https://github.com/RackHD/on-taskgraph/pull/149

PR #136, opened 2016-07-05
Race condition in task-graph can cause false success

stuart-stanley:
Issue: https://github.com/RackHD/RackHD/issues/295

This is a workaround for this issue for 1.2 only and is targeted at the one graph we use for switches that hit the race condition. All the change does is force the tasks to run in serial.

@RackHD/corecommitters @johren @derrickostertag


benbp:
👍

benbp:
@VulpesArtificem

tldavies:
👍

PR #137, opened 2016-07-08
Fix graph completion state race condition when multiple terminal task…

benbp:
…s finish

Fixes https://github.com/RackHD/RackHD/issues/295

If we succeed a terminal task, and fail another terminal task at the same time, there is a race condition where the successful task can trigger a set of the graph to a succeeded state before the failed task can trigger a set to a failed state. This is because the check for whether a graph should be set as successful only includes a check for whether it has no active tasks, rather than no active or failed tasks.

Requires https://github.com/RackHD/on-core/pull/176

This issue can be reproduced (prior to these changes) by running the following graph:

```
{ 
    "friendlyName": "Graph.testrace",
    "injectableName": "Graph.testrace",
    "options": {
        "defaults": {
            "delay": 100
        },
        "4": {
            "delay": 250
        },
        "5": {
            "delay": 250,
            "fail": true
        },
        "6": {
            "delay": 260
        }
    },
    "tasks": [
        { 
            "label": "1", 
            "taskName": "Task.noop" 
        },
        {
          "label": "2",
          "taskName": "Task.noop",
          "waitOn":
            { "1": "succeeded" }
        },
        {
          "label": "3",
          "taskName": "Task.noop",
          "waitOn":
            { "2": "succeeded" }
        },
        {
          "label": "4",
          "taskName": "Task.noop",
          "waitOn":
            { "3": "succeeded" }
        },
        {
          "label": "5",
          "taskName": "Task.noop",
          "waitOn":
            { "3": "succeeded" }
        },
        {
          "label": "6",
          "taskName": "Task.noop",
          "waitOn":
            { "3": "succeeded" }
        }
    ]
}

```

And changing the `_run` function for on-tasks/lib/jobs/noop-job.js to 
```
    NoOpJob.prototype._run = function run() {
        var self = this;
        logger.info("RUNNING NOOP JOB");
        Promise.delay(this.delay).then(function() {
            if (self.options.fail) {
                self._done(new Error("Injected failure"));
                return;
            }
            self._done();
        });
    };
```

@stuart-stanley @RackHD/corecommitters @VulpesArtificem

### DIFF
--- a/lib/completed-task-poller.js
+++ b/lib/completed-task-poller.js
@@ -151,7 +151,7 @@ function completedTaskPollerFactory(
                 data.done = true;
                 return Rx.Observable.just(data);
             }
-            return store.checkGraphFinished(data);
+            return store.checkGraphSucceeded(data);
         })
         .flatMap(function(_data) {
             if (_data.done) {
--- a/lib/task-scheduler.js
+++ b/lib/task-scheduler.js
@@ -326,7 +326,7 @@ function taskSchedulerFactory(
         var self = this;
 
         return Rx.Observable.just(data)
-        .flatMap(store.checkGraphFinished.bind(store))
+        .flatMap(store.checkGraphSucceeded.bind(store))
         .filter(function(_data) { return _data.done; })
         .flatMap(store.setGraphDone.bind(store, Constants.Task.States.Succeeded))
         .filter(function(graph) { return !_.isEmpty(graph); })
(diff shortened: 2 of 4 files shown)


pscharla:
@stuart-stanley @RackHD/corecommitters This one needs some reviews - it has been sitting for a while

stuart-stanley:
@benbp : I can't recall. Did we finally decide this does work or doesn't? I seem to recall something along the lines of while it all settles out correctly in the end, if someone came in and did a query via REST against it during the settling time, it could get the wrong answer?

stuart-stanley:
Ok, it looks like that discussion was over in https://github.com/RackHD/on-core/pull/176. As long as it feels like this will all work together correctly when in, +1.

PR #140, opened 2016-07-12
Fix a timeout issue with the centOS install graph

tannoa2:
@RackHD/corecommitters @jlongever @zyoung51 @keedya @uppalk1 @amymullins


yyscamper:
:+1:

PR #141, opened 2016-07-13
proposed appending datestring to version in bintray packages

heckj:
partial resolution for https://github.com/RackHD/RackHD/issues/310 - equiv changes would be needed for all the package repositories (on-http, on-dhcp-proxy, on-tftp, ..., etc)

Basically pulling the same scenario we used in https://github.com/RackHD/on-taskgraph/blob/master/HWIMO-BUILD to include an increasing version as we made additional builds on a branch. Date string is formatted to be increasing over time, and mapped (pretty closely) to the last commit merged into the branch.

### DIFF
--- a/extra/make-deb.sh
+++ b/extra/make-deb.sh
@@ -13,6 +13,12 @@ else
    BRANCH=${TRAVIS_BRANCH}
 fi
 
+# this appends a datestring formatting specifically to be increasing
+# based on the last date of the commit in this branch to provide increasing
+# DCH version numbers for building debian packages for bintray.
+GITCOMMITDATE=$(git show -s --pretty="format:%ci")
+DATESTRING=$(date -d "$GITCOMMITDATE" -u +"%Y-%m-%d-%H%M%SZ")
+
 if [ -z "$DEBFULLNAME" ]; then
         export DEBFULLNAME=`git log -n 1 --pretty=format:%an`
 fi
@@ -22,7 +28,7 @@ if [ -z "$DEBEMAIL" ]; then
 fi
 
 if [ -z "$DEBBRANCH" ]; then
-        export DEBBRANCH=`echo "${BRANCH}" | sed 's/[\/\_]/-/g'`
+        export DEBBRANCH=`echo "${BRANCH}${DATESTRING}" | sed 's/[\/\_]/-/g'`
 fi
 
 if [ -z "$DEBPKGVER" ]; then
@@ -55,6 +61,6 @@ debuild --no-lintian --no-tgz-check -us -uc
 popd
 if [ ! -d deb ]; then
   mkdir deb
-fi 
+fi
 
 cp -a *.deb deb/
--- a/extra/make-deb.sh
+++ b/extra/make-deb.sh
@@ -17,7 +17,7 @@ fi
 # based on the last date of the commit in this branch to provide increasing
 # DCH version numbers for building debian packages for bintray.
 GITCOMMITDATE=$(git show -s --pretty="format:%ci")
-DATESTRING=$(date -d "$GITCOMMITDATE" -u +"%Y-%m-%d-%H%M%SZ")
+DATESTRING=$(date -d "$GITCOMMITDATE" -u +"%Y%m%d%H%M%SZ")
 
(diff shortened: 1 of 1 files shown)


heckj:
@RackHD/corecommitters @ytjohn @jlmcknight Hey gang - take a quick look at this, if you like it, I'll create equiv updates to the other repositories (on-http, on-tftp, on-syslog, on-dhcp-proxy) to set their debian packages to be increasing as well.

When using bintray generated packages, I *think* this will solve the problem of being able to run something like `sudo apt-get update; sudo apt-get dist-upgrade` and get the latest code from bintray directly, since the debian packages will be getting "increasing versions"

extra/make-deb.sh
```diff
@@ -13,6 +13,12 @@ else
    BRANCH=${TRAVIS_BRANCH}
 fi
 
+# this appends a datestring formatting specifically to be increasing
+# based on the last date of the commit in this branch to provide increasing
+# DCH version numbers for building debian packages for bintray.
+GITCOMMITDATE=$(git show -s --pretty="format:%ci")
+DATESTRING=$(date -d "$GITCOMMITDATE" -u +"%Y-%m-%d-%H%M%SZ")
```
ytjohn: I would recommend leaving the dashes out of the datestring

`date -d "$GITCOMMITDATE" -u +"%Y%m%d%H%M%SZ`  resulting in something like `20160427181718Z`.

extra/make-deb.sh
```diff
@@ -22,7 +28,7 @@ if [ -z "$DEBEMAIL" ]; then
 fi
 
 if [ -z "$DEBBRANCH" ]; then
-        export DEBBRANCH=`echo "${BRANCH}" | sed 's/[\/\_]/-/g'`
+        export DEBBRANCH=`echo "${BRANCH}${DATESTRING}" | sed 's/[\/\_]/-/g'`
```
ytjohn: Maybe a dash between BRANCH and DATESTRING?

`master2016-04-27-181718Z` vs `master-2016-04-27-181718Z` vs `master-20160427181718Z`.

ytjohn:
This will be a great step to get started. That way, for any set of packages installed, it'll be easy to reference the commit date.  Some projects put a `git rev-parse HEAD` into either a VERSION or a COMMIT file that ends up with the final installation. Not saying it's needed here, just something to think about.

heckj:
@ytjohn you wanting that information (git hash) in the debian file metadata, or a file literally named `COMMIT` in the filesystem when the package is deployed? You want to have the commits of the local libraries as well (on-tasks, on-core) in there as well to know all the pieces checked out?

ytjohn:
@heckj I did some looking since my last comment, and didn't find any clear examples of putting the vcs commit id in the package.  I think when I saw it before, it would be appended to the control file as either a `VCS-Commit:` or `Description-Commit` field. Which would be visible when viewing the package information.

That being said, I'd much rather get versioned packages in now, and worry about matching to a particular git commit later if needed.

heckj:
I updated the PR and compressed the datestring and added a dash between branch name and the timestamp. The resulting debian built from this looks like:

`on-taskgraph_1.1-1master-20160714150836Z1_amd64.deb`

benbp:
👍

yyscamper:
I think @ytjohn 's suggestion is good. The date-string is not enough to specify the exact code version, but commit id can.
For this PR, it is :+1: , @ytjohn 's suggestion can be included in another PR.

anhou:
@ytjohn @yyscamper by default, debian build tool doesn't support add new a field. I have try that if set a new field such as 'Description-Commit', it will be ignored in final package. I think one place could put the commit information to the 'changelog' after format change.  need to  `apt-get changelog xxx` to get changelogs. 
for package version date is better for RackHD's daily build and no released package, so :+1: for me

@heckj  just a question that how many space for RackHD？ I see that the change mostly repos like on-http is about 50Mbytes on-taskgraph is about 20MBytes,  and I go through the travisci build history  that about 1 build per day in average, so maybe about 30 x (20MB + 50MB)  ~= 2GB after 1 month.

PR #142, opened 2016-07-15
Move version option from default to install-os task for photon os

iceiilin:
For Photon installation workflow, version and repo options are only used for "install os" task. Move it out from default so that other task would not bother the render failure problem if the payload set the install-os task only.
@yyscamper @cgx027

### DIFF
--- a/lib/graphs/install-photon-graph.js
+++ b/lib/graphs/install-photon-graph.js
@@ -9,11 +9,10 @@ module.exports = {
         defaults: {
             // Make sure this matches for both the install-os task and the
             // rackhd-callback-uri-wait task, so put it in "defaults"
-            completionUri: 'renasar-ansible.pub',
-            version: null,
-            repo: '{{api.server}}/photon/{{options.version}}'
+            completionUri: 'renasar-ansible.pub'
         },
         'install-os': {
+            version: null,
             schedulerOverrides: {
                 timeout: 3600000 // 1 hour
             }


lib/graphs/install-photon-graph.js
```diff
@@ -9,11 +9,10 @@ module.exports = {
         defaults: {
             // Make sure this matches for both the install-os task and the
             // rackhd-callback-uri-wait task, so put it in "defaults"
-            completionUri: 'renasar-ansible.pub',
-            version: null,
-            repo: '{{api.server}}/photon/{{options.version}}'
+            completionUri: 'renasar-ansible.pub'
         },
         'install-os': {
+            version: null,
```
WangWinson: It looks like version are required in install os schema and it's a `string` type.  @yyscamper Will this fail schema validation ?

yyscamper: Sure this will fail the schema, but that's what we want, because that's just tell user he must input a valid user to replace the `null`. So in theory, add to remove this line will not impact the expected behavior, remove it seems better.

But the tricky is that, this test case (https://github.com/RackHD/on-taskgraph/blob/master/spec/lib/graph-library-spec.js#L48) will try to create the graph instance if no `null` option existing, the test will obviously fail as some required option is missing. So add a `null` option is just want to make sure the former test case will not be run.

In the long run, that test case doesn't make sense, I plan to improve this test case or completed remove it.

Add @benbp for awareness.

yyscamper: I just remember I created a `validateDefinition` for task-graph (https://github.com/RackHD/on-tasks/blob/master/lib/task-graph.js#L554), we can change the unit-test to only do `validateDefinition` rather than creating the whole graph instance. That's not belong to this PR, I can submit a PR to fix this.

yyscamper:
test this please

lib/graphs/install-photon-graph.js
benbp: @yyscamper agreed, it's a hacky test case for sure and we should improve it. I think there is value in having some test case that validates everything though, as I've seen multiple times in the past the tests catching an invalid graph before merge. It's really the only way we have right now to assert that our graph library is 100% valid at the moment.

yyscamper:
test this please

lib/graphs/install-photon-graph.js
yyscamper: @benbp : I find this library (https://github.com/json-schema-faker/json-schema-faker), we can use it to generate fake graph options, so that everything is ready to create the graph instance.

yyscamper:
:+1:

WangWinson:
looks like no gating issue, right ?
:+1:

PR #146, opened 2016-07-20
Add Redfish system specific discovery graph

jlongever:
- Create active discovery graph for Redfish System resources.

@RackHD/corecommitters @zyoung51 @uppalk1 @tannoa2 @keedya

### DIFF
--- a/lib/graphs/discovery-redfish-system-graph.js
+++ b/lib/graphs/discovery-redfish-system-graph.js
@@ -1,0 +1,127 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Redfish System Discovery',
+    injectableName: 'Graph.Redfish.System.Discovery',
+    options: {
+        'bootstrap-ubuntu': {
+            'triggerGroup': 'bootstrap',
+            'kernelFile': 'vmlinuz-3.16.0',
+            'initrdFile': 'ramfs.lzma'
+        },
+        'finish-bootstrap-trigger': {
+            'triggerGroup': 'bootstrap'
+        },
+        'reboot-start': {
+            'obmServiceName': 'redfish-obm-service',
+            'force': true
+        },
+        'reboot-end': {
+            'obmServiceName': 'redfish-obm-service',
+            'force': true
+        }
+    },
+    tasks: [
+        {
+            label: 'set-boot-pxe',
+            taskName: 'Task.Obm.Node.PxeBoot',
+            ignoreFailure: true
+        },
+        {
+            label: 'reboot-start',
+            taskName: 'Task.Obm.Node.Reboot',
+            waitOn: {
+                'set-boot-pxe': 'finished'
+            }
+        },
+        {
+            label: 'bootstrap-ubuntu',
+            taskName: 'Task.Linux.Bootstrap.Ubuntu',
+            waitOn: {
+		        'reboot-start': 'finished'
+            }
+        },
+        {
+            label: 'catalog-dmi',
+            taskName: 'Task.Catalog.dmi'
+        },
+        {
+            label: 'catalog-ohai',
+            taskName: 'Task.Catalog.ohai',
+            waitOn: {
+                'catalog-dmi': 'finished'
+            }
+        },
+        {
+            label: 'catalog-bmc',
+            taskName: 'Task.Catalog.bmc',
+            waitOn: {
+                'catalog-ohai': 'finished'
+            },
+            ignoreFailure: true
+        },
+        {
+            label: 'catalog-lsall',
+            taskName: 'Task.Catalog.lsall',
+            waitOn: {
+                'catalog-bmc': 'finished'
+            },
+            ignoreFailure: true
+        },
+        {
+            label: 'catalog-megaraid',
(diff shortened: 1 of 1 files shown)


lib/graphs/discovery-system-graph.js
```diff
@@ -0,0 +1,129 @@
...
+	    'obmServiceName': 'redfish-obm-service',
+	    'force': true
+        },
+	'reboot-end': {
+	    'obmServiceName': 'redfish-obm-service',
+	    'force': true
+        },
+	'bootstrap-ubuntu': {
+	    'kernelFile': 'vmlinuz-3.16.0',
+            'initrdFile': 'ramfs.lzma'
```
anhou: indent

anhou:
@jlongever  filename  'discovery-redifish-system-graph.js' will be more accurate?  
:+1:  with minor comments update

lib/graphs/discovery-system-graph.js
```diff
@@ -0,0 +1,129 @@
...
+    friendlyName: 'Redfish System Discovery',
+    injectableName: 'Graph.System.Discovery',
+    options: {
+        'bootstrap-ubuntu': {
+            'triggerGroup': 'bootstrap'
+        },
+        'finish-bootstrap-trigger': {
+            'triggerGroup': 'bootstrap'
+        },
+	'reboot-start': {
```
yyscamper: indent problem

lib/graphs/discovery-system-graph.js
```diff
@@ -0,0 +1,129 @@
...
+            label: 'reboot-start',
+            taskName: 'Task.Obm.Node.Reboot',
+            waitOn: {
+                'set-boot-pxe': 'finished'
+            }
+        },
+        {
+            label: 'bootstrap-ubuntu',
+            taskName: 'Task.Linux.Bootstrap.Ubuntu',
+	    waitOn: {
```
yyscamper: indent problem

lib/graphs/discovery-system-graph.js
```diff
@@ -0,0 +1,129 @@
...
+		'reboot-start': 'finished'
+            }
+        },
+        {
+            label: 'catalog-dmi',
+            taskName: 'Task.Catalog.dmi'
```
yyscamper: Seems miss adding the `ignoreFailure` for dmi catalog and ohai catalog

lib/graphs/discovery-system-graph.js
```diff
@@ -0,0 +1,129 @@
...
+            taskName: 'Task.Obm.Node.Reboot',
+            waitOn: {
+                'catalog-lldp': 'finished'
+            }
+        },
+        {
+            label: 'finish-bootstrap-trigger',
+            taskName: 'Task.Trigger.Send.Finish',
+            waitOn: {
+                'reboot-end': 'finished'
```
yyscamper: If waitOn the `finished` state, the `reboot-end` task is better to be ignore failure. If don't want to ignore failure of reboot-end, you'd better change to wait on the success of 'reboot-end'

yyscamper:
jenkins: test this please

lib/graphs/discovery-system-graph.js
jlongever: Are you suggesting the normal discovery graph is wrong as well and all catalog tasks should set `ignoreFailure` to `true`?   https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/discovery-graph.js#L23 

IMHO,  dmidecode and ohai are standard to a system and _should not_ be ignored if they fail.  I actually don't even think `lshw` should even be ignored either:  https://github.com/RackHD/on-taskgraph/blob/master/lib/graphs/discovery-graph.js#L46

lib/graphs/discovery-system-graph.js
jlongever: I don't care if it succeeded, I just want to wait for it to be finished.  
I want `reboot-end` to fail, particularly in the case where testing on the current hardware has random reset problems.  This allows the workflow to fail,  signalling the user who is watching the workflow status.

jlongever:
@anhou I'm not sure if that is more accurate.  "System" with OBM service being Redfish is implied it's a Redfish system to me,  but I will change the name if you think that's better.

yyscamper:
@jlongever : There are still a lot indent problem

lib/graphs/discovery-system-graph.js
yyscamper: Make sense if you expect the reboot failure to fail the whole workflow. 
Both `success` and `finished` will work as expected, in my opinion, use `success` will be more straightforward, it explicitly shows reboot must be success. Use `finished` is not so explicit as former.
Since both behave the same , so this is not a gating problem to me.

lib/graphs/discovery-system-graph.js
yyscamper: OK, I previously thought all catalogs task should be ignore failure, seems we had intentionally to expect some standard catalog to be success, so no problem here now.

yyscamper:
@jlongever : please fix those indent problems first.

benbp:
:+1:

jlongever:
test this please

PR #147, opened 2016-07-25
Feature/use init for startup

davidzuhn:
Ubuntu 16.04 uses systemd, which has a compability mode for init scripts and not for upstart.   This patch contains a rudimentary init script (start only, matching the capabilities of the upstart script) and works on Ubuntu 14.04 *and* Ubuntu 16.04.

.gitignore has been updated to reflect the .xz files created on Ubuntu 16.


davidzuhn:
test this please

anhou:
:+1:

anhou:
test this please

davidzuhn:
@RackHD/corecommitters  please review

benbp:
FYI @zhangy42, since I believe you're working on systemd scripts for CentOS.

zhangy42:
I am working on researching how to deploy monorail on CentOS, and as a result, there will be a list of works to do for a decent deploy. There will definitely have one item in the list that is about writing systemd scripts/configuration file.

davidzuhn:
Depending on your desired CentOS target version, you may be better off using the init compatibility there as well, at least for the time being.  I suspect there's little else within RackHD that would require CentOS 7, and being able to support both 6 and 7 could be useful.

PR #150, opened 2016-07-29
Disable ACPI in kernel

jlongever:
Temporarily disable ACPI in kernel bootstrap.
Updated to cleanup tasks, replace OBM reset task with shell reboot at the end of discovery.

@RackHD/corecommitters @uppalk1 @zyoung51 @tannoa2 @keedya

### DIFF
--- a/lib/graphs/discovery-redfish-system-graph.js
+++ b/lib/graphs/discovery-redfish-system-graph.js
@@ -8,8 +8,7 @@ module.exports = {
     options: {
         'bootstrap-ubuntu': {
             'triggerGroup': 'bootstrap',
-            'kernelFile': 'vmlinuz-3.16.0',
-            'initrdFile': 'ramfs.lzma'
+            'kargs': {'acpi':'off'}
         },
         'finish-bootstrap-trigger': {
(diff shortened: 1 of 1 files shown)


keedya:
:+1:

zyoung51:
+1

heckj:
👍 , but why? 

And is this a temporary temporary, or possibly a long term fix to something?

jlongever:
@heckj hopefully temporary,  current platform-a hangs with ACPI enabled in the kernel,  this makes the normal discovery images work.

heckj:
Ah, right-o - thanks!

jlongever:
@heckj also,  shouldn't be any significant drawback to disabling ACPI,  at least in our discovery images.  Probably just lose some power capabilities that are superfluous to cataloging.

heckj:
Had to look it up again, but once I did I clued into what ACPI capabilities generally provided - seemed like primarily a channel used to let the hardware send a signal to software running on the motherboard to clean up, because it was about to be shut down.

benbp:
:+1:

PR #156, opened 2016-08-11
Better init script

ytjohn:
- enable start/stop/restart and pid monitoring
 - put logs in /var/log/rackhd

https://github.com/RackHD/on-http/issues/416

### DIFF
--- a/debian/on-taskgraph.init
+++ b/debian/on-taskgraph.init
@@ -24,6 +24,10 @@ USAGE="Usage: $0 {start|stop|restart|status} [--force]"
 FORCE_OP=false
 RESTART=false
 
+if [ -f /etc/default/$NAME ]; then
+        . /etc/default/$NAME
+fi
+
 pid_file_exists() {
     [ -f "$PID_FILE" ]
 }
@@ -38,10 +42,8 @@ is_running() {
 }
 
 start_it() {
-    if [ ! -d $PID_DIR ]; then mkdir -p "$PID_DIR";  fi
-    chown $USER:$USER "$PID_DIR"
-    mkdir -p "$LOG_DIR"
-    chown $USER:$USER "$LOG_DIR"
+    if [ ! -d $PID_DIR ]; then mkdir -p "$PID_DIR"; fi
+    if [ ! -d $LOG_DIR ]; then mkdir -p "$LOG_DIR"; fi
 
     echo "Starting $NAME ..."
     /sbin/start-stop-daemon --background --start --quiet\
@@ -116,11 +118,14 @@ status_app() {
         then
             PID=$(get_pid)
             echo "$NAME running with pid $PID"
+            exit 0
         else
             echo "$NAME stopped, but pid file exists"
+            exit 1
         fi
     else
         echo "$NAME stopped"
+        exit 3
     fi
 }
 
--- a/debian/on-taskgraph.init
+++ b/debian/on-taskgraph.init
@@ -24,9 +24,8 @@ USAGE="Usage: $0 {start|stop|restart|status} [--force]"
 FORCE_OP=false
 RESTART=false
 
-if [ -f /etc/default/$NAME ]; then
-        . /etc/default/$NAME
-fi
+# Override defaults and don't start unless default file exists.
+. /etc/default/$NAME
 
 pid_file_exists() {
     [ -f "$PID_FILE" ]
--- a/debian/on-taskgraph.init
+++ b/debian/on-taskgraph.init
@@ -24,8 +24,14 @@ USAGE="Usage: $0 {start|stop|restart|status} [--force]"
 FORCE_OP=false
 RESTART=false
 
-# Override defaults and don't start unless default file exists.
-. /etc/default/$NAME
+if [ ! -f "/etc/default/$NAME" ]
+then
+    echo "Service disabled due to absence of /etc/default/$NAME" 1>&2
+    exit 1
+else
+     . "/etc/default/$NAME"
+fi
+
 
 pid_file_exists() {
     [ -f "$PID_FILE" ]
--- a/debian/on-taskgraph.init
+++ b/debian/on-taskgraph.init
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-. /lib/lsb/init-functions
-
 ### BEGIN INIT INFO
 # Provides:          on-taskgraph
 # Required-Start:    $syslog $network mongodb rabbitmq-server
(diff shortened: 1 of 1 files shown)


ytjohn:
Hold up on this - still has issues relating to background task. Might implement start-stop-daemon.

ytjohn:
Upgraded script to use start-stop-daemon.

changev:
retest this please

changev:
retest this please

leoyjchang:
I like this enhancement.  Add @davidzuhn (original author and expert) and other people for review as well. @panpan0000 @changev @PengTian0

jimturnquist:
I loaded this on an appliance running RackHD and verified that start/stop/restart are working well.

+1

ytjohn:
I did a bit more looking, and I realized I should add exit levels.  Let me push up a commit that includes those.

https://refspecs.linuxbase.org/LSB_3.0.0/LSB-PDA/LSB-PDA/iniscrptact.html

debian/on-taskgraph.init
```diff
@@ -12,37 +10,156 @@
...
+}
+
+is_running() {
+    PID=$(get_pid)
+    ! [ -z "$(ps aux | awk '{print $2}' | grep "^$PID$")" ]
+}
+
+start_it() {
+    if [ ! -d $PID_DIR ]; then mkdir -p "$PID_DIR";  fi
+    chown $USER:$USER "$PID_DIR"
```
davidzuhn: The ownership of $PIDDIR and $LOG_DIR should only be changed or set if actually created by this script.  

Otherwise you're possibly polluting files that might be installed from other packages, and changing their contents without good reason.   All that's needed is for the current user to be able to write to $PID_DIR, not that the ownership of $PID_DIR is the current user.

debian/on-taskgraph.init
```diff
@@ -12,37 +10,156 @@
...
+    mkdir -p "$LOG_DIR"
+    chown $USER:$USER "$LOG_DIR"
+
+    echo "Starting $NAME ..."
+    /sbin/start-stop-daemon --background --start --quiet\
```
davidzuhn: A space before the backslash here is a good idea.

debian/on-taskgraph.init
```diff
@@ -12,37 +10,156 @@
...
+PID_DIR="/var/run"
+PID_FILE="$PID_DIR/$NAME.pid"
+LOG_DIR="/var/log/rackhd"
+LOG_FILE="$LOG_DIR/$NAME.log"
+NODE_EXEC=$(which node)
+
+USAGE="Usage: $0 {start|stop|restart|status} [--force]"
+FORCE_OP=false
+RESTART=false
+
```
davidzuhn: How about source /etc/default/on-taskgraph here, so that any of these variables could be changed in that script?

debian/on-taskgraph.init
```diff
@@ -12,37 +10,156 @@
...
-	/usr/bin/nodejs /var/renasar/on-taskgraph/index.js >> "$LOGFILE" 2>&1 &
-	log_success_msg "Started on-taskgraph daemon"
+        if is_running
+        then
+            PID=$(get_pid)
+            echo "$NAME already running with pid $PID"
+            exit 1
+        else
+            echo "$NAME stopped, but pid file exists"
+            if [ $FORCE_OP = true ]
```
davidzuhn: For an unattended appliance configuration, we want as few things in the way of starting up the daemons as possible.   So there really needs to be a way to enable the ''force" by default.    

Better yet would be for the daemon in question to not keep running when it can't bind to the ports it needs.  I'm not sure what use case keeps the daemon running when it can't communicate with the rest of the world.   

Then you could start the daemon without much worrying about whether another one is already running (it should be idempotent until such time as it has acquired the necessary resources, including bound network ports, to function as "the service").

debian/on-taskgraph.init
ytjohn: That's a great idea. Will add it in.

debian/on-taskgraph.init
ytjohn: Good point. Will adjust.

debian/on-taskgraph.init
```diff
@@ -1,7 +1,5 @@
 #!/bin/sh
 
-. /lib/lsb/init-functions
```
davidzuhn: So if you're concerned about things like LSB exit code compliance, why did you remove the use of log_{success,failure}_msg and replace them with echo?

ytjohn: I was primarily working off another template I had that seemed to cover most use cases: recording the pid, putting node's foreground only process into the background in a non-blocking fashion, and allowing for restarts. I discovered that it didn't have proper exit codes, which means that config management tools will assume the service is running when they check and get an exit 0.

I'm reading up on the LSB expected behavior and I can convert these echoes to lsb log messages as well.

davidzuhn: I wouldn't actually worry too much about that -- I doubt that there's a huge base of tools that expect the LSB exit codes to be meaningful.   If there is that base, I'd love to be surprised if/when you send me a URL or 2.   success/failure is all you need 99% of the time.

debian/on-taskgraph.init
```diff
@@ -12,37 +10,161 @@
...
+PID_FILE="$PID_DIR/$NAME.pid"
+LOG_DIR="/var/log/rackhd"
+LOG_FILE="$LOG_DIR/$NAME.log"
+NODE_EXEC=$(which node)
+
+USAGE="Usage: $0 {start|stop|restart|status} [--force]"
+FORCE_OP=false
+RESTART=false
+
+if [ -f /etc/default/$NAME ]; then
```
davidzuhn: Your script has a change from the existing functionality -- if the /etc/default/<service> script does not exist, the service is not started.   This version will start the service whether or not the /etc/default/<service> file is present.

This was done to ensure that services did not start up until they were properly configured (which may not happen until after the package is installed).  Since Debian/Ubuntu tries hard to start the service as the package is installed and configured, this gate on the service startup was put into place.

PengTian0:
test this please

debian/on-taskgraph.init
ytjohn: Is this still a valid concern? I think anyone that is familiar with services knows to restart the service after they modify the configuration. Almost every major application will install and start, even if the defaults aren't very useful.

jimturnquist:
What is left for this to be merged?

ytjohn:
I will go ahead and revert to requiring /etc/default/$NAME in order to get this in place.

ytjohn:
@davidzuhn  On each PR, I have set it back to require /etc/default/$NAME

```
Stopping on-http ...
Killing process 30085
Removing pid file
on-http stopped
root@rackhd01:~# rm /etc/default/on-http 
root@rackhd01:~# /etc/init.d/on-http start
/etc/init.d/on-http: 28: .: Can't open /etc/default/on-http
root@rackhd01:~# touch /etc/default/on-http
root@rackhd01:~# /etc/init.d/on-http start
Starting on-http ...
on-http started with pid 2284
```

All changes are pushed. Let me know if there's anything else preventing these from being merged in.

rackhdmirror:
Build finished.

debian/on-taskgraph.init
```diff
@@ -12,37 +10,160 @@
...
+LOG_DIR="/var/log/rackhd"
+LOG_FILE="$LOG_DIR/$NAME.log"
+NODE_EXEC=$(which node)
+
+USAGE="Usage: $0 {start|stop|restart|status} [--force]"
+FORCE_OP=false
+RESTART=false
+
+# Override defaults and don't start unless default file exists.
+. /etc/default/$NAME
```
davidzuhn: I would change wrap this in a conditional:
```
if [ -f "/etc/default/$NAME" ]
then
    . "/etc/default/$NAME"
fi
```

which makes the unusual message ```/etc/init.d/on-http: 28: .: Can't open /etc/default/on-http``` go away.  That message sort of looks like an error, but it's not really.

davidzuhn:
I think ```/etc/init.d/on-http: 28: .: Can't open /etc/default/on-http``` is a not-very useful message.   

```
if [ ! -f "/etc/default/$NAME" ]
then
    echo "Service disabled due to absence of /etc/default/$NAME" 1>&2
    exit 1
else
     . "/etc/default/$NAME"
fi

**Comment by davidzuhn** [comment]
In any case, this looks functional as is.   +1

File: debian/on-taskgraph.init (hunk 7 above)
Comment (ytjohn) [inline debian/on-taskgraph.init]: That is what I had earlier this week. You said  9 days agot that you wanted to require /etc/default/$NAME in order to prevent the service from starting up before the service was configured.

https://github.com/RackHD/on-taskgraph/pull/156/files/b4816d9c6adc4b71f0f48e3df5af2519d4519dbf#r76984313

**Comment by ytjohn** [comment]
Added error message suggested by @davidzuhn 

```
root@rackhd01:~# /etc/init.d/on-taskgraph start
Service disabled due to absence of /etc/default/on-taskgraph
root@rackhd01:~# touch /etc/default/on-taskgraph
root@rackhd01:~# /etc/init.d/on-taskgraph start
on-taskgraph already running with pid 30132
```

jimturnquist:
Can we get this one merged? Is there anything left that is needed?

yyscamper:
:+1:

PR #157, opened 2016-08-12
docker: Add symlink to ipmitool

codenrhoden:
When installing ipmitool on Alpine, it does not go to /usr/bin/ipmitool,
it's in /usr/sbin. As a workaround, add a symlink to it until
RackHD is doing proper $PATH evaluation.

See: https://github.com/RackHD/RackHD/issues/155

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,8 @@ RUN mkdir -p ./node_modules \
   && ln -s /RackHD/on-core/node_modules/di ./node_modules/di \
   && npm install --ignore-scripts --production \
   && echo "@testing http://nl.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories \
-  && apk add --update ipmitool@testing net-snmp net-snmp-libs net-snmp-tools
+  && apk add --update ipmitool@testing net-snmp net-snmp-libs net-snmp-tools \
+  && ln -s /usr/sbin/ipmitool /usr/bin/ipmitool
 
 VOLUME /var/lib/dhcp
 CMD [ "node", "/RackHD/on-taskgraph/index.js" ]


heckj:
:+1: for the workaround until we get a "proper solution" into our underlying code for RackHD/RackHD#155

yyscamper:
I add some enhancement in this PR: RackHD/on-tasks#315

My solution is to check the existence of some required tool at the early stage when service starts. This depends on user has correctly added the tool in the $PATH.

For the new change, RackHD will use `ipmitool` rather than `usr/bin/ipmitool`.

yyscamper:
Sorry, I made a mistake to close this PR, but I have re-open it. ^^

yyscamper:
jenkins: test this please

brianparry:
test this please

anhou:
@codenrhoden could you help confirm if @yyscamper 's PR https://github.com/RackHD/on-tasks/pull/315  resolve your issue, if yes, this PR could be closed

brianparry:
@anhou I don't think https://github.com/RackHD/on-tasks/pull/315 will fix this issue so much as find it earlier.  I think this workaround is still needed.

codenrhoden:
@brianparry @anhou I think RackHD/on-tasks#315 actually does fix the issue, since it changes the calls to `ipmitool` to be just `ipmitool` instead of `/usr/bin/ipmitool`. It was always the hardcoded path that was the problem, not that the tool wasn't installed. If you agree, I think you can indeed close this in deference to the other PR.

This was meant to be a workaround until something like that PR came around. Little did I know @yyscamper would do it in two days! 😀

yyscamper:
I see @brianparry has some concern about the upfront checking of required tools, I agree those code is better to belong to the job and it's job's responsibility to check the tool existence.

I will move the checkRequiredTools from task-service into each job and also keep the change the use `ipmitool` rather than `/user/bin/ipmitool`.

changev:
Jenkins: test this please

changev:
Jenkins: test this please

changev:
Jenkins: test this please

changev:
Jenkins: test this please

changev:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

changev:
test this please

codenrhoden:
@yyscamper Yep!  Thanks for the heads up.

PR #181, opened 2016-10-31
increase ssh-validation timeout to 60 sec for CentOS install

geoff-reid:
Increased ssh-validation timeout to fix failing CentOS 7.0 max payload install.  
@keedya @uppalk1 @tannoa2 @RackHD/corecommitters


geoff-reid:
test this please

PR #183, opened 2016-11-04
remove ssh validation from OS installation workflow

lanchongyizu:
Remove ssh validation from OS installation workflow since its funtionality for alerting the completion of OS intallation can be replaced with notification.

@RackHD/corecommitters @panpan0000 @iceiilin @johren @keedya


keedya:
@lanchongyizu  thanks for this changes. We still need to change RackHD os-install regression tests and remove anything in the payload related to ssh validation

lanchongyizu:
@anhou https://github.com/RackHD/RackHD/pull/486 is for tests, and there needs no changes to docs.
Thanks.

PR #189, opened 2016-11-25
Remove unecessary graphId info for progress

pengz1:
Jenkins depends on: https://github.com/RackHD/on-core/pull/227, https://github.com/RackHD/on-http/pull/523, https://github.com/RackHD/on-core/pull/227, https://github.com/RackHD/on-tasks/pull/370
This PR is to remove unecessary graphId in task start/finish progress event.

### DIFF
--- a/lib/task-runner.js
+++ b/lib/task-runner.js
@@ -16,8 +16,7 @@ di.annotate(taskRunnerFactory,
         'Rx',
         'Task.Task',
         'Task.Messenger',
-        'TaskGraph.Store',
-        'TaskGraph.TaskGraph'
+        'TaskGraph.Store'
     )
 );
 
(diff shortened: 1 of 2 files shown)


pengz1:
@RackHD/corecommitters @cgx027 @iceiilin

anhou:
@pengz1 Any background for this?

pengz1:
@anhou GraphId is already includes in progress info, I used to included graphId in taskProgress data because when we call store.updateTaskProgress, graphId is required inside this method. Later I changed store.updateTaskProgress parameter list form (data) to (data, graphId) so that we don't have to include graphId inside data. After this change, graphId is redundant in taskProgress (which is the real "data" that be delivered to store.updateTaskProgress).

cgx027:
:+1:

PR #191, opened 2016-12-13
Fix periodic race condition in task runner heartbeat test

benbp:
Resolves https://github.com/RackHD/RackHD/issues/351

The issue with this test is that the Rx take operator will limit the number of times the `subscribe` callback is called, but not the number of times the methods in the observable pipeline are called, and the test checks the callcount of a stub called in the pipeline. The reason we see the race condition is sometimes the test completes the callcount check fast enough before the next iteration of the 1ms interval we use, and sometimes it doesn't

To understand the Rx behavior, consider the difference between this code:

```
var run = true;

Rx.Observable.interval(1)
.takeWhile(function() {
    return run;
})
.tap(function() {
    console.log('tap');
})
.subscribe(
    function() { run = false; },
    function() {},
    function() { console.log('complete'); }
);
```

Which outputs
```
$ node test.js
tap
complete
```

versus this code:

```
var run = true;

Rx.Observable.interval(1)
.tap(function() {
    console.log('tap');
})
.takeWhile(function() {
    return run;
})
.subscribe(
    function() { run = false; },
    function() {},
    function() { console.log('complete'); }
);
```

Which outputs
```
$ node test.js
tap
tap
```

@RackHD/corecommitters @tldavies @VulpesArtificem @heckj

### DIFF
--- a/spec/lib/task-runner-spec.js
+++ b/spec/lib/task-runner-spec.js
@@ -256,8 +256,8 @@ describe("Task Runner", function() {
 
         it('should heartbeat Tasks on an interval', function(done) {
             runner.running = true;
-            runner.heartbeat = Rx.Observable.interval(1);
-            var heartStream = runner.createHeartbeatSubscription(runner.heartbeat).take(5);
+            runner.heartbeat = Rx.Observable.interval(1).take(5);
+            var heartStream = runner.createHeartbeatSubscription(runner.heartbeat);
             streamOnCompletedWrapper(heartStream, done, function() {
                 expect(store.heartbeatTasksForRunner.callCount).to.equal(5);
             });


heckj:
+1

benbp:
To answer a question from @heckj, the effective pipeline that gets built up includes what's returned from `runner.createHeartbeatSubscription`:

```
        return  heartInterval
                .takeWhile(self.isRunning.bind(self))
                .flatMap(store.heartbeatTasksForRunner.bind(store, self.taskRunnerId))
                .flatMap( function(taskCount) {
                    if(taskCount < Object.keys(self.activeTasks).length){
                        return self.handleUnownedTasks();
                    } else if (taskCount > Object.keys(self.activeTasks).length) {
                        return self.handleLostTasks();
                    }
                    return Rx.Observable.just(null);
                })
                .catch(function(error) {
                    logger.error('Failed to update heartbeat, stopping task runner and tasks', {
                        taskRunnerId: self.taskRunnerId,
                        error: error,
                        activeTasks: _.keys(self.activeTasks)
                    });
                    return Rx.Observable.just(self.stop.bind(self)());
                });
```

Currently, we add `take()` to the END of that chain, after the `heartbeatTasksForRunner` stub. This change adds it to the START, so it becomes, effectively:

```
heartInterval
.take(5)
// rest of stuff
```

PR #193, opened 2016-12-20
Add set-bmc task to Discovery

keedya:
- Make set obm settings optional and configurable
-  by adding option obmOption: true in rackhd config file

### DIFF
--- a/lib/graphs/discovery-graph.js
+++ b/lib/graphs/discovery-graph.js
@@ -11,6 +11,10 @@ module.exports = {
         },
         'finish-bootstrap-trigger': {
             'triggerGroup': 'bootstrap'
+        },
+        'obm-option' : {
+            obmOption: 'false',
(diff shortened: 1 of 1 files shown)


keedya:

lib/graphs/discovery-graph.js
```diff
@@ -11,6 +11,10 @@ module.exports = {
         },
         'finish-bootstrap-trigger': {
             'triggerGroup': 'bootstrap'
+        },
+        'obm-option' : {
+            obmOption: 'false',
```
yyscamper: The variable name `obmOption` isn't so clear for me to reflect its purpose, I suggest choosing another better name. For example, "autoCreateObm" or you have better choice.

keedya: @yyscamper I didn't have a chance to change the name. I will make the change in a different PR

PR #195, opened 2016-12-22
Update decommission compute node workflow

keedya:
- Fix the order of tasks


keedya:

PR #196, opened 2016-12-27
add timeout value for cancelTaskGraph

iceiilin:
It is found on Jenkins continues test that sometimes on-taskgraph abnormally exits right after starting (about 1 times per day). The error msg is
```
12:52:27 PM graph.1 |  Request Timed Out (on.task-graph-runner:methods.cancelTaskGraph:Object).
12:52:27 PM graph.1 |  RequestTimedOutError: Request Timed Out (on.task-graph-runner:methods.cancelTaskGraph:Object).
12:52:27 PM graph.1 |      at null.<anonymous> (/home/vagrant/src/on-taskgraph/node_modules/on-core/lib/common/messenger.js:361:17)
12:52:27 PM graph.1 |      at Timer.listOnTimeout (timers.js:92:15)
12:52:27 PM graph.1 Exited Abnormally
```
It happened only below two conditions exist at the same time:
1. service graph (poller workflow) in mongoDB is different from current code, so that cancelTaskGraph needs to be called right after RackHD starts to clear previous workflow.
2. when in taskgraph scheduler, one promise is processing the cancel request, while either of the streams (evaluateTask, evaluateGraph or checkGraphFinish) are scheduled to run, which costs more time for the previous promise to finish. 

The default timeout is 1s for all requests in RackHD, and this PR modifies it to 5s for this request. Jenkins continues test has run this patch for 1 week, and the above problem doesn't occur any more.

Theoretically, it is still possible that the timeout value is not enough, especially in scale environment, and service graph has changed after last run. But we could not wait endlessly, neither could we ignore the timeout error, because the code will delete corresponding mongoDB document after cancelling graph.

Jenkins depends on: https://github.com/RackHD/on-core/pull/232

### DIFF
--- a/lib/service-graph.js
+++ b/lib/service-graph.js
@@ -17,7 +17,10 @@ di.annotate(ServiceGraph, new di.Inject(
     )
 );
 function ServiceGraph(TaskGraph, store, taskGraphProtocol, Constants, Promise, _) {
-    var exports = {};
+    var exports = {
+        // Set cancelTaskGraph default timeout
+        timeout: 5000
+    };
 
     /**
      * @param {Object} definition - the JSON graph definition that the TaskGraph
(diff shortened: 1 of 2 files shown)


iceiilin:
jenkins: test this please

changev:
jenkins depends on: https://github.com/RackHD/on-core/pull/232
test this please

PR #198, opened 2017-01-06
add vscode config into gitignore list

yyscamper:
Add vscode config file into gitignore list since I am using vscode for remote debugging, it is annoying to see it in "git status".


PengTian0:
test this please

PR #200, opened 2017-01-10
Add an option to make pollers creation on compute node optional

keedya:
- In the future, the same option can be used for other node types


keedya:

PR #202, opened 2017-01-12
fix houndci parse error by removing comment in jshintrc

yyscamper:
The houndci doesn't strip the comment from the jshintrc file (a JSON file), so it complains an error about "fail to parsing config file".

It is true that the standard JSON doesn't support comment, so my fix is to completely remove the comment from jshintrc file.

@anhou @iceiilin @pengz1


pengz1:
+1

PR #204, opened 2017-01-13
Unit test fixes

dalebremner:
@brianparry @srinia6 @BillyAbildgaard


brianparry:
FYI - I expect jenkins to fail since this feature branch does not exist in all the other repositories.

PR #210, opened 2017-02-03
Bugfix: for set bmc credentials

keedya:
- Migrate set bmc credentials from discovery graph to Discovery-sku-graph
  in discovery graph, group label is set in bootstrap
  and it overides the tasks properties

fix for : https://github.com/RackHD/RackHD/issues/602

depends on : https://github.com/RackHD/on-http/pull/578


keedya:
@manfrednde @brianparry @geoff-reid @tannoa2 @uppalk1

PR #213, opened 2017-02-08
Fix Discovery graph

keedya:
Fix Discovery failure introduced by adding set bmc credentials as an option


keedya:
@geoff-reid @brianparry @uppalk1 @tannoa2 @dalebremner @BillyAbildgaard @srinia6

PR #216, opened 2017-02-17
 Update secure erase graph

pengz1:
Jenkins depends on : https://github.com/RackHD/on-tasks/pull/414, https://github.com/RackHD/on-tasks/pull/415, https://github.com/RackHD/on-tasks/pull/408
Secure erase graph is updated based on two considerations:
1. To avoid wrong disk info input by user. If user doesn't use RackHD to do RAID operation, it might happen that user input disk device name doesn't match disk info in existing RackHD catalog. In this case secure erase workflow could erase a disk that user doesn't want to erase.
2. To add delay between OS ready and drives ready. After boot into microkernel, it takes something for OS to scan all disks in the case large quantity of disks are installed on the node. A secure erase task or drive re-cataloging tasks immediately after OS is ready may fail since only part of disks are scanned.

Solutions in the new graph:
1. For the first problem, driveId catalogs will be cached, also driveId and megaraid catalogs will be refreshed before secure erase and compared with cached driveId catalog to identify if  raid configuration is changed. Secure erase will not be executed if catalogs are not identical.
2. For the second problem, a new task is added before re-cataloging tasks, which will delay OS for a user-configurable duration after node boots into microkernel.

### DIFF
--- a/lib/graphs/secure-erase-drive-graph.js
+++ b/lib/graphs/secure-erase-drive-graph.js
@@ -7,11 +7,14 @@ module.exports = {
     injectableName: 'Graph.Drive.SecureErase',
     options: {
         'bootstrap-ubuntu': {
-            overlayfsFile: "secure.erase.overlay.cpio.gz",
+            overlayfsFile: 'secure.erase.overlay.cpio.gz',
             triggerGroup: 'secureErase'
         },
         'drive-secure-erase': {
             eraseSettings: null
+        },
+        'drive-scan-delay': {
+            duration: 10000
         },
         'finish-bootstrap-trigger': {
             triggerGroup: 'secureErase'
@@ -19,9 +22,16 @@ module.exports = {
     },
     tasks: [
         {
+            label: 'cache-driveId-catalog',
+            taskName: 'Task.Get.DriveId.Catalog',
+        },
+        {
             label: 'set-boot-pxe',
             taskName: 'Task.Obm.Node.PxeBoot',
-            ignoreFailure: true
+            ignoreFailure: true,
+            waitOn: {
+                'cache-driveId-catalog': 'succeeded'
+            }
         },
         {
             label: 'reboot',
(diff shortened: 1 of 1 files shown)


pengz1:
@iceiilin @sunnyqianzhang @lanchongyizu

iceiilin:
+1

lib/graphs/secure-erase-drive-graph.js
```diff
@@ -38,14 +38,22 @@ module.exports = {
             }
         },
         {
+            label: 'catalog-megaraid-before',
+            taskName: 'Task.Catalog.megaraid',
+            waitOn: {
+                'reboot': 'succeeded'
+            },
+            ignoreFailure: true
```
lanchongyizu: Since secure erase is based on Megaraid catalog, ignoreFailure shouldn't be true.
@yyscamper , please confirm this.

lanchongyizu:
:+1:

lib/graphs/secure-erase-drive-graph.js
yyscamper: I agree @lanchongyizu 's comment

lib/graphs/secure-erase-drive-graph.js
```diff
@@ -38,14 +38,22 @@ module.exports = {
             }
         },
         {
+            label: 'catalog-megaraid-before',
```
yyscamper: The `driveId` is still need be refreshed before running secure erase.

pengz1: Refresh driveId catalog may introduce inconsistency on drive info if user did RAID operation. This might lead to disk matching error and wrong secure erase operation.

lib/graphs/secure-erase-drive-graph.js
pengz1: We should depend on user configured Megaraid catalog not this one. This one is just a backup and this workflow is not intend to catalog Megaraid information.

sunnyqianzhang:
:+1:

lib/graphs/secure-erase-drive-graph.js
```diff
@@ -7,21 +7,31 @@ module.exports = {
...
+        'drive-scan-delay': {
+            duration: 10000
+        },
         'finish-bootstrap-trigger': {
             triggerGroup: 'secureErase'
         }
     },
     tasks: [
         {
+            label: 'cache-driveId-catalog',
```
brianparry: @pengz1 Hi Krein, how does this task work before `set-boot-pxe` and `reboot`?

brianparry:
@pengz1 Need to make note that this PR depends on https://github.com/RackHD/on-tasks/pull/415.  Also, it seems like we're missing a test for this graph in our FIT/CIT test suite since testing passes without this dependency.

lib/graphs/secure-erase-drive-graph.js
```diff
@@ -38,33 +48,54 @@ module.exports = {
             }
         },
         {
+            label: 'drive-scan-delay',
```
brianparry: @pengz1 I'm curious why you chose to implement a new sleep task which sleeps the micro-kernel shell rather than using a noop task and delay option.

pengz1: There is different in secure erase since linux bootstrap is started with secure erase tasks (or re-cataloging tasks in the new graph). A noop task will  start from last reboot and thus I need to calculate time from reboot to OS ready pluses time from OS ready to disks ready.
For a node delay, it means we only needs delay time for OS ready to disks ready and we can skip the reboot time which varies on platforms and BIOS conditions.

lib/graphs/secure-erase-drive-graph.js
pengz1: It is just caching existing catalogs into graph context, any timing is OK before re-cataloging

PR #224, opened 2017-03-10
[skip ci]change latest master tag from lastest to devel

changev:


panpan0000:
👍 
background: RAC-3820

PR #226, opened 2017-03-16
filter the empty data when graph not found

VulpesArtificem:
part of the fix for https://rackhd.atlassian.net/browse/RAC-637 
supports https://github.com/RackHD/on-core/pull/269
@pscharla @RackHD/corecommitters 

If we couldn't get graph the task belongs to it's probably because the graphobjects collection was cleared we're not going to be able to get it, just filter the undefined.


yyscamper:
jenkins: test this please

PR #227, opened 2017-03-19
Update DMI catalog after firmware update

dalebremner:


pengz1:
Update Dell server DMI catalog will introduce booting microkernel to the previously OBM workflow, this will consume more time to the workflow.
A better suggestion is to move this workflow to in-band mode if you do need specified firmware version here. However to create a in-band workflow for Dell servers we need CentOS microkernel with the latest RachalOS microkernel.
Another suggestion is to get all firmware version information as an independent catalog, this is more reasonable and can provide user more information.
I am fine with the workflow itself, but I can see we are going to modify this workflow in the future.

dalebremner:
Hi @pengz1. Thanks for the comments. Correct, this is a short term solution, which will be replaced by using the Dell micro-service for upgrading firmware. Eventually, we will collect all firmware versions, as you suggest...

PR #230, opened 2017-03-24
Fix: graph progress is incorrect when graph failed, timeout or cancelled

lanchongyizu:
The original issue is that  graph progress is incorrect when graph failed, timeout or cancelled.

The root cause is that the method `tap` in Rxjs is not waited to finish, and the data flowing into tap is modified by later stream operation.

The solution is `cloneDeep` the `object` to be published.

Recently, https://github.com/RackHD/on-taskgraph/pull/278 has `cloneDeeped` the `graph`, so the problem has been solved in fact. So this updated PR is to `cloneDeep` the remain `objects` to be published.

@RackHD/corecommitters @pengz1 @iceiilin

### DIFF
--- a/lib/task-runner.js
+++ b/lib/task-runner.js
@@ -339,7 +339,10 @@ function taskRunnerFactory(
                 );
             })
             .tap(function(task) {
-                return graphProgressService.publishTaskStarted(task, {swallowError: true});
+                //The reason to cloneDeep the task is that the method tap in Rxjs is not waited
+                //to finish, and the data flowing into tap is modified by later stream operation.
+                var taskCopy = _.cloneDeep(task);
+                return graphProgressService.publishTaskStarted(taskCopy, {swallowError: true});
             })
             .tap(function(task) {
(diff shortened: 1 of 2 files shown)


lib/task-runner.js
```diff
@@ -339,7 +339,8 @@ function taskRunnerFactory(
                 );
             })
             .tap(function(task) {
-                return graphProgressService.publishTaskStarted(task, {swallowError: true});
+                var taskCopy = _.cloneDeep(task);
+                return graphProgressService.publishTaskStarted(taskCopy, {swallowError: true});
```
yyscamper: Why the graphProgressService need to change the input task object? I am think moving the deeply clone to inside graphProgressService will be better.

lanchongyizu: @yyscamper graphProgressService doesn't change the input task object, but it may be changed by the later stream operations.

pengz1:
You should explain why deep copy is required in your comments

PR #235, opened 2017-04-17
Allow scripts to run in npm install

keedya:
- ignore-scripts flag applies to all the installed packages.
In this case , part of grps installation is to oompile C code
- requireWrapper default home directory is different in docker
  using singleWrapper instead


keedya:
@johren @brianparry @dalebremner @tannoa2

PR #239, opened 2017-04-26
bugfix: add on-tasks files copy in docker build

changev:
Add on-tasks files copy in docker build.
Otherwise on-taskgraph will always use the latest on-tasks.

https://rackhd.atlassian.net/browse/RAC-4956

@panpan0000 @PengTian0


PengTian0:
:+1:

PR #245, opened 2017-05-10
on-taskgraph code

longhornteam35:
on-taskgraph code additions. Includes graphs for create volume, delete
volume, and add hotspare.


iceiilin:
I would suggest to move dell specific tasks and workflows into on-skupack as other platforms, instead of putting into rackhd directly. Users could add these features dynamically by posting skupack to rackhd if using dell server.

PR #257, opened 2017-07-06
 Remove legacy callback options for run-emc-diag

pengz1:
Remove legacy options in run-emc-diag-graph.js. Callback has already been replaced by notification API.


pengz1:
@nortonluo @iceiilin

PR #258, opened 2017-07-11
RAC-4843

BrentHiggins:
* Adding cleanup field to the dell-wsman-update-systemcomponents-graph that will remove the specified file if true.


BrentHiggins:
@jimturnquist @derrickostertag

PR #260, opened 2017-07-18
[RAC-5468]on-taskgraph docker image's latest version in DockerHub are not getting updated correctly

HQebupt:
In order to figure out who when where build a docker image and push to the dockerhub.
Related:
- [RackHD/on-tasks/pull/479](https://github.com/RackHD/on-tasks/pull/479)
- [RackHD/on-http/pull/700](https://github.com/RackHD/on-http/pull/700)

**NOTICE**: It will modify the docker file in order to add more information into the docker image which is not expected in production environment.
**After solving the issue, I will discard the changes.**
@changev @panpan0000 @PengTian0 @iceiilin @anhou


build_debug/build_image_info.sh
```diff
@@ -0,0 +1,17 @@
...
+set -x
+
+cd $(dirname $0)
+
+tty=$(who)
+hostname=$(hostname)
+ip_info=$(ifconfig | grep "inet addr")
+os_info=$(lsb_release -a)
+
+echo -e "tty:\n${tty}\n\nhostname:\n${hostname}\n\nip_info:\n${ip_info}\n\nos_info:\n${os_info}" | openssl rsautl -encrypt -pubin -inkey build_image_key.pub > build_image_info.encrypted
```
anhou: I think 'tty' is useless, and 'who' should be the user, like 'jenkins@vmslave', and also I didn't see 'when'

build_debug/build_image_info.sh
```diff
@@ -0,0 +1,20 @@
...
+# Purpose: figure out RAC-5468. 
+# If you want to decrypt the file "build_image_info.encrypted", please contact Qiang, Peter, Alan, Leo.
+
+set -x
+
+cur_time=$(date)
+cur_user=$USER
+cur_path=$(pwd)
+
+user_login=$(who)
```
panpan0000: any diff than cur_user?

build_debug/build_image_info.sh
```diff
@@ -0,0 +1,20 @@
...
+hostname=$(hostname)
+ip_info=$(ifconfig | grep "inet addr")
+os_info=$(lsb_release -a)
+
+cd $(dirname $0)
+echo -e "current time:${cur_time}\ncurrent user:${cur_user}\ncurrent path:${cur_path}\n\nuser login:\n${user_login}\n\nhostname:\n${hostname}\n\nip info:\n${ip_info}\n\nos info:\n${os_info}" | openssl rsautl -encrypt -pubin -inkey build_image_key.pub > build_image_info.encrypted 
+
```
panpan0000: BTW: I'm thinking , can we just dump the whole "ENV" ?

build_debug/build_image_info.sh
HQebupt: "who" displays who is logged on to the system. The output result is more than one user probably.
If Alan and I login on the same machine both , there are two users with the command "who".

build_debug/build_image_info.sh
HQebupt: The whole "ENV" has redundant information, so we can grep the useful info.

changev:
+1 after pasting test results

changev:
docker build run in docker containers 

so all the the env vars of hosts can't be got in docker images~

Please read docker doc and find a way to pass host envs to docker containers.

HQebupt:
The script in docker file is running in the docker environment, not in the host. So the PR could not help to solve the issue.

PR #261, opened 2017-07-19
 Modify run-emc-diag workflow

pengz1:
Backgroup
Current run-emc-diag workflow is old and not working during our test.

Details
 1. Use force pxe boot task and obm reset to make sure boot flag won't
 be cleared by emc platform rebooting
 2. Add more retries to accommodate emc platform long time BMC
 inaccessibility
 3. Add unit test for the graph.

@nortonluo @iceiilin @anhou

### DIFF
--- a/lib/graphs/run-emc-diag-graph.js
+++ b/lib/graphs/run-emc-diag-graph.js
@@ -5,16 +5,25 @@
 module.exports = {
     friendlyName: 'Run EMC Diagnostics',
     injectableName: 'Graph.Run.Emc.Diag',
-    options: {},
+    options: {
+        'set-boot-pxe': {
+            "delay": 1000,
+            "retries": 10
+        },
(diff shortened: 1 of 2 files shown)


lib/graphs/run-emc-diag-graph.js
```diff
@@ -5,16 +5,25 @@
...
         {
             label: 'set-boot-pxe',
-            taskName: 'Task.Obm.Node.PxeBoot',
+            taskName: 'Task.Obm.Force.Pxe.Boot',
             ignoreFailure: true
         },
         {
             label: 'reboot-start',
-            taskName: 'Task.Obm.Node.Reboot',
+            taskName: 'Task.Obm.Node.Reset',
```
iceiilin: Why changing from reboot to reset?

PR #264, opened 2017-08-02
Make taskgraph stand alone service: working callback

tannoa2:
Signed-off-by: Geoff Reid <geoff.reid@dell.com>


tannoa2:
jenkins depends on: https://github.com/RackHD/on-http/pull/723
jenkins depends on: https://github.com/RackHD/RackHD/pull/848

api/rest/notification.js
```diff
@@ -0,0 +1,37 @@
+// Copyright 2017, EMC Inc.
+
+'use strict';
+
+var injector = require('../../index.js').injector;
+var controller = injector.get('Http.Services.Swagger').controller;
+var notificationApiService = injector.get('Http.Services.Api.Notification');
```
iceiilin: This name is the same as that in on-http, which might confuse others as whether these two processes have dependencies. I would suggest to change it to some name like "Taskgraph.Services.Api.Notification", and it could also be applied to other services under on-taskgraph.

anhou:
@tannoa2   @geoff-reid  Before going into the details. I have a question that how do you plan to treat southbound/northbound in on-http service, and related concept in on-taskgraph.  or Both on-http and on-taskgraph have nortbound and southbound? how do they communicate? I also raise the same question previously. it was not answered clearly due to it was in POC stage then. now it's submitting the codes, I think we have to understand clearly before they're merged

keedya:
@RackHD/committers

app.js
```diff
@@ -0,0 +1,177 @@
...
+            server.listen(config.httpPort, config.hostname, function () {
+                logger.info('Your server is listening on port %d'.format(config.httpPort));
+                logger.info('Swagger-ui is available on http://%s:%d/docs'
+                    .format(config.hostname, config.httpPort));
+            });
+        });
+    }
+
+    function stop() {
+
```
keedya: @tannoa2 this is missing

`function stop() {
        return Promise.map(services, function(service) {
             return service.stop();
        })
        .then(function() {
            return core.stop();
        });
    }`

index.js
```diff
@@ -112,20 +72,7 @@ taskGraphRunner.start(options)
...
-        .finally(function () {
-            if (server) {
-                server.close();
-            }
-            process.nextTick(function () {
-                process.exit(1);
-            });
-        });
-});
+
```
keedya: @tannoa2 we need this code for SIG interrupts

PR #267, opened 2017-08-08
RAC-5886 Update dell-wsman-update-systemcomponents-graph.js to get la…

patelb10:
…test inventory


patelb10:
It was noticed that attributes were not applied when updating via:
/redfish/v1/Systems/{id}/Bios/Settings
The workflow was working although the inventory was never requeried.
Here we update "dell-wsman-update-systemcomponents-graph.js" so it can get new updated attributes when complete.

patelb10:
Paired with @jimturnquist

patelb10:
@johren

PR #271, opened 2017-08-11
RAC-5886 Update dell-wsman-update-systemcomponents-graph.js to get la…

patelb10:
…test inventory


patelb10:
@jimturnquist @pathakvaidehi2391

PR #278, opened 2017-08-24
To updte the failed graph status before pulbishing graph fail

sunnyqianzhang:
Before code change, when a taskgraph failed, it will first publish the graph finished event, and then update the taskgraph status. The issue it introduced is that: for the published graph, the status is not yet up-to-date.

The code change is to fix the problem: when a graph failed, change the graph status before publishing the graph finished event.

@pengz1 @iceiilin @lanchongyizu 
Note this code review and the other unmerged code review all changed the file lib/task-schedule.js, whichever one is merged first, the other two need rebase.
 https://github.com/RackHD/on-taskgraph/pull/
https://github.com/RackHD/on-taskgraph/pull/230

### DIFF
--- a/lib/task-scheduler.js
+++ b/lib/task-scheduler.js
@@ -412,17 +412,12 @@ function taskSchedulerFactory(
      */
     TaskScheduler.prototype.failGraph = function(data, graphState) {
         var self = this;
+        var graphToBePublished;
         return Rx.Observable.just(data.graphId)
         .flatMap(store.getActiveGraphById)
         .filter(function(graph) {return !_.isEmpty(graph);})
-        .tap(function(graph) {
-            return graphProgressService.publishGraphFinished(
-                graph,
-                graphState,
-                {swallowError: true}
-            );
-        })
         .map(function(doneGraph) {
+            graphToBePublished = _.cloneDeep(doneGraph);
             return _.map(doneGraph.tasks, function(taskObj) {
                 if(taskObj.state ===  Constants.Task.States.Pending) {
                     taskObj.state = graphState;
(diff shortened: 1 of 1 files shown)


lib/task-scheduler.js
```diff
@@ -428,7 +429,6 @@ function taskSchedulerFactory(
             });
         })
         .flatMap(self.handleFailGraphTasks.bind(self))
-        .flatMap(store.setGraphDone.bind(store, graphState, data))
         // setGraphDone can return null if another source has already updated
         // the graph state. Don't publish the same event twice.
         .filter(function(graph) { return graph; })
```
lanchongyizu: This filter is related with `setGraphDone,` so we can't just move `setGraphDone` only. 
In addition, since we want to `publish progress finished event` after graph is set to Done, we can keep a copy of original graph data, and `publish progress finished event` after the filter.

lib/task-scheduler.js
```diff
@@ -409,6 +409,7 @@ function taskSchedulerFactory(
         var self = this;
         return Rx.Observable.just(data.graphId)
         .flatMap(store.getActiveGraphById)
+        .flatMap(store.setGraphDone.bind(store, graphState, data))
```
pengz1: You should at least move this line after the "filter" line.

lib/task-scheduler.js
pengz1: Agree

lib/task-scheduler.js
```diff
@@ -412,17 +412,13 @@ function taskSchedulerFactory(
      */
     TaskScheduler.prototype.failGraph = function(data, graphState) {
         var self = this;
+        var graphToBePublished;
         return Rx.Observable.just(data.graphId)
         .flatMap(store.getActiveGraphById)
+        //.flatMap(store.setGraphDone.bind(store, graphState, data))
```
lanchongyizu: No need to add this line.

lib/task-scheduler.js
```diff
@@ -433,15 +429,22 @@ function taskSchedulerFactory(
             });
         })
         .flatMap(self.handleFailGraphTasks.bind(self))
-        .flatMap(store.setGraphDone.bind(store, graphState, data))
         // setGraphDone can return null if another source has already updated
         // the graph state. Don't publish the same event twice.
+        .flatMap(store.setGraphDone.bind(store, graphState, data))
```
lanchongyizu: No need to move this line.

PR #280, opened 2017-08-31
Fix unit test error caused by modification on on-core

sunnyqianzhang:
jenkins: depends on https://github.com/RackHD/on-core/pull/296
The error found in continuous-integration/travis-ci/pr is expected and could be ignore. When pulling the two dependent PR down and run locally, the issue doesn't exists


sunnyqianzhang:
test this please

sunnyqianzhang:
The error found in continuous-integration/travis-ci/pr is expected and could be ignore. When pulling the two dependent PR down and run locally, the issue doesn't exists

PR #292, opened 2017-10-05
fix a buffer limit in the southbound

keedya:


keedya:
@geoff-reid  @tannoa2

PR #316, opened 2017-10-31
Remove obsolete parameter after WSMAN control job refine

nortonluo:
Background

The original WSMAN based job is out of date after SMI service updated its input parameters. And also it is lack of unit test and option schema check. Currently we refine all WSMAN implementation and aimed to integrate SMI service with RackHD and replace the RACADM. We refined the WSMAN work job arch. Now it use the new common WSMAN base job as base classs and WSMAN tool as util to send out request to SMI service. The parameters used in old wsman taskgraph are obsolete now.

Detail

Remove obsolete parameters in firmware update graph.

Test Result

Tested pass on Dell R730 server.

Review: @yaolingling @anhou @lanchongyizu @AlaricChan

### DIFF
--- a/lib/graphs/dell-wsman-update-firmware-graph.js
+++ b/lib/graphs/dell-wsman-update-firmware-graph.js
@@ -5,13 +5,7 @@
 module.exports = {
     friendlyName: 'Dell wsman Update Firmware Graph',
     injectableName: 'Graph.Dell.Wsman.Update.Firmware',
-    options: {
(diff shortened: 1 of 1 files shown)


lanchongyizu:
Please add `run-test` label to trigger Concourse CI.

nortonluo:
@lanchongyizu I thinks it's better to relay on schema.  The default value has no meaning(NULL) here to user.

PR #321, opened 2017-11-02
Update graph to use task

mtannous:
get-switch-config-graph.js
get-switch-version-graph.js
update-switch-firmware-graph.js


keedya:
@mtannous  does this PR has a on-tasks dependency?

mtannous:
Yes, it is in this one : https://github.com/RackHD/on-tasks/pull/542

nucklehead:
@keedya I think this can be merged now :-)

PR #327, opened 2017-11-09
Fix graph parameter

nortonluo:
## Background
This PR will fix the npm test fail in on-taskgraph. The unit test in on-taskgraph will check the paramters in schema. If the parameter do not have default value, it will throw an error. Recently the schema of wsman firmware update is newly updated, and will result unit test fail.

## Details
Set all the parameters required by schema in on-tasks default value as null.

### DIFF
--- a/lib/graphs/dell-wsman-update-firmware-graph.js
+++ b/lib/graphs/dell-wsman-update-firmware-graph.js
@@ -5,7 +5,11 @@
 module.exports = {
     friendlyName: 'Dell wsman Update Firmware Graph',
     injectableName: 'Graph.Dell.Wsman.Update.Firmware',
-    options: {},
+    options: {
+        shareFolderAddress: null,
+        shareFolderType: null,
+        shareFolderName: null
+    },
     tasks: [
         {
             label: 'dell-wsman-update-firmware',


yyscamper:
This is not about missing default value, but I believe this is a trick point. 

The unit-test only picks those graph definitions that don't have "null" options, so if you set a "null" option, the unit-test for that graph definition is actually been skipped. This code tell you everything: https://github.com/RackHD/on-taskgraph/blob/master/spec/lib/graph-library-spec.js#L53 .

Since you set a 'null' option, you are trying to skip the unit-test for that graph.

The unit-test for task-definition has the similar problem before, I fixed it by leveraging the task schema to generate faked options (thanks to the library [json-schema-faker](https://github.com/json-schema-faker/json-schema-faker). For example, if option `host` is required, it should be an IP address, but it doesn't have a default value in task definition, then `json-schema-faker` will use the schema of option `host` to generate a faked IP address (apparently it can pass the task schema validation). 
You can find the code for task definition validation from here: https://github.com/RackHD/on-tasks/blob/master/spec/lib/task-data/tasks/base-tasks-spec.js#L99

The benefit is:
(1) If the task has options schema defined, then it can be validated no matter those required options has default value or not.
(2) If the task has default options, then it can validate whether the default options is correct or not.

The reason why not implement this to graph definition:
(1) there are still many tasks don't have schema defined. But a graph depends on all embedded tasks have schema.

anhou:
this issue is fixed  by another PR https://github.com/RackHD/on-taskgraph/pull/328

PR #328, opened 2017-11-09
fix unit test for wsman update FW graph

keedya:


keedya:
@geoff-reid @uppalk1 @tannoa2

keedya:
@dalebremner

PR #329, opened 2017-11-09
Create a Discovery sku graph that use old mikrokernel

keedya:
- This option can be configured in the rackhd config.json file
- to use it, user can set property "discoveryGraph": "Graph.SKU.Orig.Discovery"


keedya:
@geoff-reid @tannoa2

keedya:
@dalebremner @king-jam

keedya:
@dalebremner to review

PR #332, opened 2017-11-10
Move to rancherOS v1.1.0 and add netconsole option

keedya:


keedya:
@geoff-reid @dalebremner @tannoa2 tests are performed in the on-imagebuilder pr referenced here

PR #333, opened 2017-11-15
add dell-wsman-reset-components-graph

lanchongyizu:
**Background**
Based on the previous investigation result, develop to port Graph.Dell.Racadm.ResetComponents and /Systems/{identifier}/Bios.ResetBios to use WSMAN API.
https://rackhd.atlassian.net/browse/RAC-6481

**Done**
add dell-wsman-reset-components-graph

@RackHD/corecommitters @nortonluo

### DIFF
--- a/lib/graphs/dell-wsman-reset-components-graph.js
+++ b/lib/graphs/dell-wsman-reset-components-graph.js
@@ -1,0 +1,20 @@
+// Copyright 2017, Dell EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Dell Wsman Reset Components Graph',
+    injectableName: 'Graph.Dell.Wsman.Reset.Components',
+    options: {
+        defaults: {
+            components: null
+        }
+    },
+    tasks: [
+        {
+            label: 'dell-wsman-reset-components',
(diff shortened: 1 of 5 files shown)


lib/graphs/dell-wsman-reset-components-graph.js
```diff
@@ -0,0 +1,20 @@
+// Copyright 2017, Dell EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Dell Wsman Reset Components Graph',
+    injectableName: 'Graph.Dell.Wsman.Reset.Components',
+    options: {
+        defaults: {
+            components: null
```
nortonluo: Can this parameter be accepted in tasks?

lanchongyizu:
depends on: https://github.com/RackHD/on-tasks/pull/557

lib/graphs/dell-wsman-reset-components-graph.js
lanchongyizu: Yes, see https://github.com/RackHD/on-http/pull/813/files?diff=unified#diff-060bec0b1b4f6db402d32c27899e5bb9R544

PR #342, opened 2017-12-05
[RAC-6385] Fix ubuntu16.04 support

mcgG:
****Background:**** 
----
Workflow cannot finish after successfully installed ubuntu16.04, this is because callback script in 16.04 cannot be executed during the init boot.

Detail:
----
Ubuntu 16.04 uses _**systemd**_ to start the boot scripts not _**init**_ in 14.04, but scripts in **_/init.d_** can still work in 16.04 because 16.04 uses _**systemd-sysv-generator**_ to generate services for each script in /init.d(purpose of backward compatibility). And there is no runlevel concept in 16.04, boot facilities should run in parallel, if sequential is required, use  **_LSB comment header_** in scripts.

For more detail about LSB please see [https://wiki.debian.org/LSBInitScripts](https://wiki.debian.org/LSBInitScripts)

@pengz1 @anhou @PengTian0 @Ara4Sh

### DIFF
--- a/data/templates/install-ubuntu/ubuntu-preseed
+++ b/data/templates/install-ubuntu/ubuntu-preseed
@@ -1,3 +1,4 @@
+### Copyright 2015-2017, Dell EMC, Inc.
 #### Contents of the preconfiguration file (for squeeze)
 ### Localization
 d-i debian-installer/locale string en_US.UTF-8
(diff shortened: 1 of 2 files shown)


RackHDCI:
- data/templates/install-ubuntu/ubuntu-preseed - data/templates/install-ubuntu/ubuntu.rackhdcallback

data/templates/install-ubuntu/ubuntu-preseed
```diff
@@ -230,6 +231,7 @@ in-target wget http://<%=server%>:<%=port%>/api/current/templates/post-install-u
 in-target chmod +x ./post-install.sh; \
 in-target ./post-install.sh; \
 in-target wget http://<%=server%>:<%=port%>/api/current/templates/ubuntu-sources?nodeId=<%=nodeId%> -O /etc/apt/sources.list; \
-in-target wget http://<%=server%>:<%=port%>/api/current/templates/<%=rackhdCallbackScript%>?nodeId=<%=nodeId%> -O /etc/rc2.d/S99callback; \
-in-target chmod +x /etc/rc2.d/S99callback; \
+in-target wget http://<%=server%>:<%=port%>/api/current/templates/<%=rackhdCallbackScript%>?nodeId=<%=nodeId%> -O /etc/init.d/S99callback; \
```
yyscamper: this change will impact both ubuntu 14.04 and 16.04, please confirm this works for both. Otherwise, you need version checking.

markbeierl:
Looks like this is still not working as the /usr/sbin/update-rc.d script is explicitly looking for only 3 ###, whereas this change uses 4.

`        if (m/^### BEGIN INIT INFO\s*$/) {
        elsif (m/^### END INIT INFO\s*$/) {
`

PR #345, opened 2017-12-09
Adding new graph worflows for Hypervisor deployment via out of band c…

rahmanmuhamad:
…ontroller , this is the initial pull request and will be followed by few more on other modules

### DIFF
--- a/lib/graphs/dell-wsman-os-deployment-create-graph.js
+++ b/lib/graphs/dell-wsman-os-deployment-create-graph.js
@@ -1,0 +1,29 @@
+// Copyright 2016, EMC, Inc.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Dell wsman OS Deployment Create ISO Graph',
+    injectableName: 'Graph.Dell.Wsman.Os.Create',
+    options: {
+        defaults: {
+           destinationDir: null,
+           destinationFileName: null,
+           fileName: null,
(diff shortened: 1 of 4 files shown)


lib/graphs/dell-wsman-os-deployment-create-graph.js
```diff
@@ -0,0 +1,29 @@
+// Copyright 2016, EMC, Inc.
```
lanchongyizu: //Copyright 2017, Dell EMC, Inc.  The same problem exists in dell-wsman-os-deployment-deploy-graph.js.

lib/graphs/dell-wsman-os-deployment-create-graph.js
```diff
@@ -0,0 +1,29 @@
...
+         {
+            label: 'dell-wsman-os-create',
+            taskName: 'Task.Dell.Wsman.Os.Create'
+        }
+    ]
+};
+
+
+
+
```
lanchongyizu: Please remove the empty lines. The same problem exists in dell-wsman-os-deployment-deploy-graph.js.

lanchongyizu:
@rahmanmuhamad Could you please help to add the unit test for both graphs?

nortonluo:
This graph depends on https://github.com/RackHD/on-tasks/pull/570

PR #355, opened 2017-12-20
[RAC-6492] Create graph to add hotspare via WSMAN

AlaricChan:
**Background**
This story is to implement adding hot spare via WSMAN, replace the RACADM dependencies in Redfish API with WSMAN implementation.

**Details**
1. create new graph
2. add UT

**Dependency**
depends on: https://github.com/RackHD/on-tasks/pull/576

**Reviewer**
@yaolingling @pengz1 @nortonluo

### DIFF
--- a/lib/graphs/dell-wsman-add-hotspare-graph.js
+++ b/lib/graphs/dell-wsman-add-hotspare-graph.js
@@ -1,0 +1,45 @@
+// Copyright © 2017 Dell Inc. or its subsidiaries. All Rights Reserved.
+
+'use strict';
+
+module.exports = {
+    friendlyName: 'Dell Wsman Add Hotspare',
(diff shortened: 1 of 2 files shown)


lib/graphs/dell-wsman-add-hotspare-graph.js
```diff
@@ -0,0 +1,46 @@
...
+
+module.exports = {
+    friendlyName: 'Dell Wsman Add Hotspare',
+    injectableName: 'Graph.Dell.Wsman.Add.Hotspare',
+    options: {
+        defaults: {
+            username: null,
+            password: null,
+            volumeId: null,
+            driveId: null,
```
yaolingling: the type of driveId is string which is defined in schema

nortonluo: DriveId default value should be empty.

PR #357, opened 2017-12-29
fix bug in megaraid-config.sh

lanchongyizu:
Graph.Bootstrap.Megaraid.Configure failed to create VD on Dell Servers with perccli64 when the drives are online.
https://rackhd.atlassian.net/browse/RAC-6583

The root cause is that we use storcli64 to find the ID of the virtual disk when the drives are online.

The solution is to replace `storcli64 /c0` with `$path /c$controller`.

@RackHD/corecommitters @pengz1


RackHDCI:
- data/templates/megaraid-config.sh

PR #359, opened 2018-01-02
Skip network VLAN configuration when list is empty

sushilrai:
Need to skip VLAN configuration when VLAN information in network device hash is empty. Currently we are only checking for null, but in case variable is defined but its empty then no network configuration is getting added to the kickstart file.

### DIFF
--- a/data/templates/centos-ks
+++ b/data/templates/centos-ks
@@ -218,7 +218,7 @@ export PATH
       interface=`grep -i /sys/class/net/*/address -e  $interface | cut -d "/" -f 5`
     <%_ } _%>
     <% if( undefined != n.ipv4 ) { %>
-      <% if( undefined != n.ipv4.vlanIds ) { %>
+      <% if( undefined != n.ipv4.vlanIds && n.ipv4.vlanIds.length > 0 ) { %>
         <% n.ipv4.vlanIds.forEach(function(vid) { %>
           echo "Configuring vlan <%=vid%> on $interface"
           sed -i '/^BOOTPROTO=/d' /etc/sysconfig/network-scripts/ifcfg-$interface.<%=vid%>


data/templates/centos-ks
```diff
@@ -218,7 +218,7 @@ export PATH
       interface=`grep -i /sys/class/net/*/address -e  $interface | cut -d "/" -f 5`
     <%_ } _%>
     <% if( undefined != n.ipv4 ) { %>
-      <% if( undefined != n.ipv4.vlanIds ) { %>
+      <% if( undefined != n.ipv4.vlanIds && n.ipv4.vlanIds.length > 0 ) { %>
```
gavin-scott: @sushilrai you can work around by not specifying the `vlanIds` parameter at all right? Like here: https://github.com/RackHD/RackHD/blob/42321cd4684d355a53f441ebaadf7ba1f52213d8/example/samples/install_centos_7_payload_full.json#L28-L32 

I think this change makes sense but it looks like all of the other kickstart templates do the same thing as this one, i.e. they don't check the length and do vlan-specific logic if `vlanIds` is set. Ideally they should all have the same behavior.

It's a little odd to me that `vlanIds` is an array at all in this context -- why would you want to have multiple interfaces on the same device with different vlans that all have the same ip address? The windows install file only ever uses the first vlan: https://github.com/RackHD/on-taskgraph/blob/31362fabd27bc1bccf987373525dead858e36333/data/templates/post-windows-install.ps1#L22

yyscamper:
Another solution (for me, it's better) is to add `"minItems": 1` restriction in task options schema, prevent user from inputing empty `vlanIds`. The task options schema validation is executed at very early stage,  this can avoid writing complex guarding logic in kickstart file.

gavin-scott:
I like that too @yyscamper , nice idea.

sushilrai:
Closing the PR as this is not consistent with other templates. "null" check is sufficient for validation. Fixed the client which was generating the request where skipped passing the VLAN IDs if it is not required to be configured

PR #360, opened 2018-01-04
RHEL 7 Bond VLAN interface support

sushilrai:
Added support for creating Tagged Bond Interface. This will enable support for having one or multiple tagged sub-interfaces created along with the LACP bond interface

### DIFF
--- a/data/templates/centos-ks
+++ b/data/templates/centos-ks
@@ -171,19 +171,14 @@ export PATH
      echo NM_CONTROLLED=no >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
      echo BOOTPROTO=none >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
      echo ONBOOT=yes >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     echo IPADDR="<%=n.ipv4.ipAddr%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     echo NETMASK="<%=n.ipv4.netmask%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     <% if ( undefined != n.ipv4.gateway) { %>
-          echo GATEWAY="<%=n.ipv4.gateway%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     <% } %>
-     echo DEFROUTE=yes >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     echo PEERDNS=yes >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-
-     <% if ( undefined != n.ipv4.dns1) { %>
-        echo DNS1="<%=n.ipv4.dns1%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     <% } %>
-     <% if ( undefined != n.ipv4.dns2) { %>
-        echo DNS2="<%=n.ipv4.dns2%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
+     <% if ( typeof n.ipv4 != 'undefined' ) { %>
+          echo IPADDR="<%=n.ipv4.ipAddr%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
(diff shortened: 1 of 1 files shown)


data/templates/centos-ks
```diff
@@ -171,19 +171,21 @@ export PATH
...
+          <% } %>
+          echo DEFROUTE=yes >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
+          echo PEERDNS=yes >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
 
-     <% if ( undefined != n.ipv4.dns1) { %>
-        echo DNS1="<%=n.ipv4.dns1%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     <% } %>
-     <% if ( undefined != n.ipv4.dns2) { %>
-        echo DNS2="<%=n.ipv4.dns2%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
+          <% if ( undefined != n.ipv4.dns1) { %>
```
yyscamper: Since the `additionalProperties` is set to false for `Ipv4Configuration` in task schema (see https://github.com/sushilrai/on-tasks/blob/b0f8e14a61f8adbc83738c3229cb6716902f3148/lib/task-data/schemas/types-installos.json#L123), so the `dns1` and `dsn2` will always be undefined if code successfully runs here. If you really need this option, then you need to add the new option into the options schema.

yyscamper:
Please also update the example payload and doc:
- https://github.com/RackHD/RackHD/blob/master/example/samples/install_rhel_payload_full.json
- http://rackhd.readthedocs.io/en/latest/rackhd/install_os.html?highlight=Install%20OS#non-windows-os-installation-workflow-payload

sushilrai:
Supported PRs
- https://github.com/RackHD/docs/pull/420
- https://github.com/RackHD/RackHD/pull/921

yyscamper:
@sushilrai good, thanks.

lanchongyizu:
@sushilrai Please help to push a new commit to trigger the concourse-ci. There are some bugs in concourse-ci. Thanks.

data/templates/centos-ks
```diff
@@ -171,19 +171,14 @@ export PATH
...
      echo ONBOOT=yes >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     echo IPADDR="<%=n.ipv4.ipAddr%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     echo NETMASK="<%=n.ipv4.netmask%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     <% if ( undefined != n.ipv4.gateway) { %>
-          echo GATEWAY="<%=n.ipv4.gateway%>" >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     <% } %>
-     echo DEFROUTE=yes >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-     echo PEERDNS=yes >> /etc/sysconfig/network-scripts/ifcfg-<%=n.name%>
-
-     <% if ( undefined != n.ipv4.dns1) { %>
```
lanchongyizu: Any reason to remove the `dns` settings?

lanchongyizu:
Hi @sushilrai, we have already supported Vlans in https://github.com/RackHD/on-taskgraph/blob/master/data/templates/centos-ks#L221-L233 and https://github.com/RackHD/on-taskgraph/blob/master/data/templates/centos-ks#L256-L268

I think your PR is duplicated with the above functionality.

sushilrai:
Hi @lanchongyizu changes in https://github.com/RackHD/on-taskgraph/pull/360/files#diff-93745c7016488bb7482deacfde1bcebfR200 are for Bonded interface specifically.

lanchongyizu:
Hi @sushilrai The original vlans settings are for bonded interfaces. Please look into the links above, the codes are almost the same.

sushilrai:
Hi @lanchongyizu content is almost the same. But the existing code is for `networkDevices`. the newly added content is for `bondvlanInterface`

https://gist.github.com/sushilrai/26b00197b232ee848d50271c40399fe4#file-rhel-rackhd-payload-json-L18-L28

lanchongyizu:
Hi @sushilrai, I thought that `bonding interface` is a special network interface, and they have a lot in common. Is it possible that we could reuse or refine existing script to adapt to both cases. But since your have defined `bonds` previously, I am fine with this `bondvlanInterface`. Maybe we could refactor it later.

PR #363, opened 2018-01-22
 Fix CentOS bug with minimal package image

pengz1:
CentOS minimal package image doesn't have wget package, this will lead
 to installation pending on requiring user confirmation and callback
 failure. To fix this:
 1. Use --ignoremissing for %packages in centos-ks.
 2. Replace all wget in centos-ks and centos callback script to curl.

@lanchongyizu @mcgG @nortonluo @HQebupt @yaolingling @AlaricChan

### DIFF
--- a/data/templates/centos.rackhdcallback
+++ b/data/templates/centos.rackhdcallback
@@ -1,4 +1,5 @@
 #! /bin/bash
+# Copyright © 2018 Dell Inc. or its subsidiaries. All Rights Reserved.
 #
 # centos.rackhdcallback       callback to rackhd post installation API hook
 #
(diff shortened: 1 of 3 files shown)


data/templates/centos.rackhdcallback
```diff
@@ -26,6 +27,7 @@
 # Nothing wrong with set -e here since we're not doing anything complex
 set -e
 echo "Attempting to call back to RackHD CentOS installer"
-wget --retry-connrefused --waitretry=1 -t 300 --post-data '{"nodeId":"<%=nodeId%>"}' --header='Content-Type:application/json' http://<%=server%>:<%=port%>/api/current/notification
+curl -X POST -H 'Content-Type:application/json'  --max-time 1000 --connect-timeout 9 --retry-delay 1 --retry 100 'http://<%=server%>:<%=port%>/api/current/notification?nodeId=<%=nodeId%>'
```
yyscamper: I remembered there is no `curl` during certain stage of installation, so used the `wget` instead. Please verify it on a lower version of CentOS, such as 6.5.

pengz1: @yyscamper  Good point. I guess this is why concourse-ci failed.
@lanchongyizu Looks we should wrap some retry mechanism here. Some packages doesn't include wget while others may don't include curl.

pengz1: Any way I will close this PR, @lanchongyizu will submit another PR.

PR #365, opened 2018-01-24
Separate ssh-job into ssh-command and ssh-catalog

lanchongyizu:
**Background**

This is to fix "No parser" issue:
```
No parser exists for command sudo ./megaraid-config.sh
No parser exists for command sudo python set_interfaces.py 
No parser exists for command sudo ipmitool chassis bootdev pxe
```

https://rackhd.atlassian.net/browse/RAC-6579

**Done**

* Separate ssh-job into ssh-command and ssh-catalog

depends on: https://github.com/RackHD/on-tasks/pull/580

@RackHD/corecommitters @pengz1 @nortonluo

### DIFF
--- a/lib/graphs/esxcli-driver-version-commands-graph.js
+++ b/lib/graphs/esxcli-driver-version-commands-graph.js
@@ -19,7 +19,7 @@ module.exports = {
             taskDefinition: {
                 injectableName: "Task.Run.Ssh",
                 friendlyName: "Ssh and run Esxi commands",
-                implementsTask: 'Task.Base.Ssh',
+                implementsTask: 'Task.Base.Ssh.Catalog',
                 options: {
                     commands: null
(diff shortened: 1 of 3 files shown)


lib/graphs/esxcli-driver-version-commands-graph.js
```diff
@@ -18,8 +18,8 @@ module.exports = {
             label: 'versionRetrieval',
             taskDefinition: {
                 injectableName: "Task.Run.Ssh",
-                friendlyName: "Ssh and run Esxi commands",
-                implementsTask: 'Task.Base.Ssh',
+                friendlyName: "Ssh and run Esxi commands, and catalog the version",
```
pengz1: I don't think it is a good idea to use a sentence as a name.

PR #366, opened 2018-01-30
Fix Debian/Ubuntu kernel panic on EFI

Ara4Sh:
Fixing kernel panic error during EFI boot on debian and ubuntu installation . 

`kernel panic not syncing vfs unable to mount root fs on unknown-block (1,0)`


Ara4Sh:
@lanchongyizu rebase is already done ;)

lanchongyizu:
@Ara4Sh Any documentation or reference about `initrd=initrd.gz` during EFI boot?

nortonluo:
Could refer to this link. This related to a kernel bug with iPXE
http://forum.ipxe.org/showthread.php?tid=7756&pid=11699#pid11699

PR #377, opened 2018-04-23
Enable Nos installation on the Dell switch s6000

keedya:
@markbeierl @dalebremner @geoff-reid


RackHDCI:
- data/templates/dell-switch-nos-install.sh

PR #380, opened 2018-05-17
add refresh-catalog tasks after creating raid successfully

lkx007:
After creating raid successfully, "driveid" catalog and "lsall" catalog needs to refresh.


iceiilin:
+1 after concourse pass the PR gate test. Currently concourse is not available

PR #383, opened 2018-05-29
Error analyzing Windows unattend file when specifying two network devices

iceiilin:
Windows installer throws error about failure to handle unsttend.xml when specifying more than one network device.
Change the order in the template from
```
<component>
    <Interfaces>
        <Interface></Interface>
    </Interfaces>
    <Interfaces>
        <Interface></Interface>
    </Interfaces>
</component>
```
into 
```
<component>
    <Interfaces>
        <Interface></Interface>
        <Interface></Interface>
    </Interfaces>
</component>
```
The latter is the same with https://docs.microsoft.com/en-us/windows-hardware/customize/desktop/unattend/microsoft-windows-tcpip-interfaces.


iceiilin:
verified pass in baremental

iceiilin:
There is no unit test for this template, so the drop in coverage could be ignored.

iceiilin:
@pengz1 @lanchongyizu please help to review

PR #384, opened 2018-05-29
[RAC-6753]Correct several taskgraph's friendlyName

leoyjchang:
*Background*
It has been found during web ui development that some graphs have extra space in their friendlyname. After further screening, I also found some typo in some graph's friendlyname.  This is not a good style and we need to fix them all.

@lanchongyizu @pengz1 @yaolingling @iceiilin @mcgG @AlaricChan

### DIFF
--- a/lib/graphs/bootstrap-ubuntu-graph.js
+++ b/lib/graphs/bootstrap-ubuntu-graph.js
@@ -3,7 +3,7 @@
 'use strict';
 
 module.exports = {
-    friendlyName: 'DEBUG Bootstrap Ubuntu ',
+    friendlyName: 'DEBUG Bootstrap Ubuntu',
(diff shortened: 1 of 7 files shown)


pengz1:
@Leo, there are some other graphs have the same friendly name, is this inside your scope?
Like rancher-discovery-graph.js vs  discovery-graph.js

leoyjchang:
@pengz1 Good point. I intended to cover duplicated friendlyname case but still miss some. I will update the PR for the case you mentioned.

pengz1:
A simple tip for you, you can use web-ui for name sorting...

pengz1:
Here is the web-ui demo link: http://10.62.59.160:8081

RackHDCI:
- lib/graphs/brocade-switch-discovery-graph.js

